PR #30, opened 2017-03-23
typo

kailas29:
Hi, I am Vasilis,

This is a test, made to see how the whole thing works

Corrected a few typos/spaces

Cheers

### DIFF
--- a/atchem.f90
+++ b/atchem.f90
@@ -459,12 +459,12 @@ PROGRAM ATCHEM
      ! printing concentration of two first species - seems unnecessary at the moment
      ! WRITE (stderr, fmt) t, y (1), y (2)
 
-     ! OUTPUT RATES OF PRODIUCTION ON LOSS (OUTPUT FREQUENCY SET IN MODEL.PARAMETERS)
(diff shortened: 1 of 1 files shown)


spco:
Hi Vasilis - great to have you onboard. Hopefully you can see above that the Travis CI build test has passed? I'll wait to hear back that you've seen how that is working, and then I'll hit merge on your first pull request!

kailas29:
Yes it gave no errors.. Is that test the same or similar to the one that is going to be running in order to verify that everything works after every change?

spco:
Congrats, your first pull request is done!
It will probably help if you now make sure your local repository is up-to-date with this one, by doing 'git pull upstream master' followed by 'git push origin master'. As you submitted from your master branch, this probably is unnecessary, but it would be if you were submitting from another branch.

Yes, that's the test (or part of the test) to check that everything works after every change. If you create a pull request, it will automatically run. If you push to an already open pull request, it will similarly run again on the newly submitted code.

Currently, the test setup is very sparse - it just checks that the program builds, and runs, and exits with zero.

Ideally, we would have a whole suite of tests that run - a couple of different compilers to check they all build correctly, plus a set of tests that check the functionality: given a set of input files, check that the output files match what we expect. That kind of thing. I need to leave for the train right now, but we can discuss what sort of thing you might be able to add to that testing regime.

PR #31, opened 2017-03-23
Neaten up some functions

spco:
Rename outputInterestingNames to outputSpeciesOutputRequiredNames. Rewrite matchNameToNumber(). Document readSpeciesOutputRequired().


spco:
So far, this also changes all `np` in atchem.f90 to `numSpec` as it is holding the number of species. Also changes a lot of occurrences of `neq` to `numSpec` as that's more appropriate.

PR #33, opened 2017-03-24
Add linux and osx as os options to .travis.yml.

spco:


spco:
This checks the build under OSX+gfortran.

PR #36, opened 2017-03-24
Add testing regime for sample inputs and outputs.

spco:


spco:
Notes to self upon return: first I need to exstract the output directory location, so I can have a full subdirectory in travis/tests/ dedicated to each of the .fac files. Output location needs to be an input parameter, based upon the name of the test (perhaps based upon the name of the .fac file, since then Python can hardcode something for now). 

Then, in this PR I can add a full comparison of all output files, not just the .out, to fully capture all output.

spco:
Note: Latest still doesn't fully handle the error/no error very well, possibly just carrying on. Needs further testing.

spco:
This completes the first stage of CI tester implementation. It includes changes to constraintFunctions.f90 to give default values to some variables which are undefined at output time.

This tests only based on a number of input .fac files, checking that they give the expected outputs (up to a numdiff tolerance) on linux and OSX.

There are improvements to make to this still, see issues #46, #47, #48, #49, #50

PR #40, opened 2017-04-03
Correct the compilation broken by PR 39.

spco:
@rs028 Foolishly I pressed 'merge' on your PR before it had finished running Travis tests. In fact, it wouldn't compile on my Mac. This should fix that issue. I was seeing `duplicate symbol` errors for each of the functions in utilityFunctions.f90.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ makefile.local:
 
 include makefile.local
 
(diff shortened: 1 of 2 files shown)


rs028:
But doesn't that mean that utilityFunction.f90 is not going to be compiled in the executable?

spco:
As `tools/mech_converter.py` wrote 
`INCLUDE 'utilityFunctions.f90'`
to `mechanism-rates.f90`, the functionality of `utilityFunctions.f90` was there, but in a slightly odd manner. I've changed this fix so the file is now compiled separately in the Makefile stage rather than using INCLUDE. This is more sane. The other two uses of INCLUDE are OK to leave for now, as they don't define proper Fortran files on their own, and those files will be disappearing soon anyway :)

PR #52, opened 2017-04-06
First attempt to create homebrew cache.

spco:


spco:
Problems encountered so far:

cache: directories doesn't seem to work properly, as the formulae are not linked. Attempts to link them required checking the right versions were being used, but then this still fails. Seems difficult to get around the brew update, unless we just nuke the /usr/local/Cellar and replace with exactly the bits we need, properly linked.

spco:
I never got this working, so I'm going to close. No point having this hanging around!

PR #53, opened 2017-04-06
Change instantaneousRates output directory to be in a user-defined position.

spco:
Need to untangle this change from the one where this is used in testing.


spco:
This changes the instantaneousRates directory to be user-input (second arg to `./athchem`), diverts test output of this type to a subdirectory of the test folder (rather than originally where it all went to the same folder in the source files) and tests this output against reference files.

It also fixes the numdiff exitcode-handling to be more robust.

PR #68, opened 2017-04-11
Envvar

rs028:
Changed a number of variables names and fixed a few typos. This is in preparation for a consistent treatment of the envVar variables.


spco:
Look good. I will squash these commits on GitHub once it's passed. Can I ask if there's a specific reason to go with PRESS rather than PRESSURE? Generally it's helpful to have the full name, rather than a contraction, now that we aren't limited to 8-character strings or anything. If anything I'd be inclined to change TEMP everywhere to TEMPERATURE, if it makes it more readable.

spco:
Looks like constraintFunctions.f90 line 195 should use pressure rather than PRESS. TEMP and M on that line should also be lowercase. If you make those changes in a new commit on the same branch, and git push, then it will update this PR appropriately.

rs028:
No specific reason apart from personal preference. I usually find it easier to understand the code if the variable name is not a proper word. But I don't have strong feelings about that, so I am happy to roll it back and change TEMP instead.

But I don't understand why it should be PRESS in that subroutine. I think it has been changed everywhere.

rs028:
Ah okay. yeah that sounds good ( maybe we should have a "style" page on the wiki :) )
I will probably get to it tomorrow, though.

rs028:
Sorry! I am running into trouble again :(

git pull upstream master
git push origin master
git checkout envvar
git rebase master

then I changed the code to fix the above issue, commit and tried 

git push origin envvar

but is is being rejected "because the tip of your current branch is behind"
I thought rebase was supposed to solve that?

spco:
Hi Roberto, can you show me the output of 

git log --branches --decorate --graph --oneline

? I expect the answer is to use git push --force, but I'd like to see that output first.

rs028:
these are the first lines

* 509360c (HEAD, envvar) Rename pressure variables
* 509360c (HEAD, envvar) Rename pressure variables
* 1e14b9c Consistent naming of envVar settings
* 9711ebc Change directory name
* 8ef2c60 Fix typos
* 2ef5fd7 Change variable name PRESSURE to PRESS
* 7a8a6c9 Consistent name of environment variables
* a27c8d8 (origin/master, origin/HEAD, master) Correct (and simplify) outputRate
* 0da10d8 Change matchNameToNumber() to output returnArraySize correctly, and re
* 726e18b Make intent lowercase.

spco:
Ok, that generally looks fine.

git push --force

should be sufficient.

rs028:
Okay, seems to be working. Thanks!

spco:
I can see this isn't going to compile because line 195 of constraintFunctions.f90 still contains a reference to PRESS which should be pressure (that's what the variable name is declared as at the top of `getEnvVarsAtT`.)

Did this compile on your machine?

rs028:
I does not compile but that's because it cannot find -lblas and -llapack
Problem is if I modify Makefile everytime I update my fork is it going to be overwritten?

spco:
Yes, it will be currently. If you have BLAS and LAPACK installed, you should be able to edit makefile.local to set LIBDIR to include the paths to LAPACK and BLAS (the basic Makefile looks through /usr/lib/ and $CVODELIB), and makefile.local will not be overwritten by git updates.

Something like this might work (both the BLASDIR and LAPACKDIR paths should end in /lib):

    BLASDIR=/path/to/blas
    LAPACKDIR=/path/to/lapack
    LIBDIR=$LIBDIR:$BLASDIR:$LAPACKDIR

This is also assuming that you have a CVODE installation working with the same LAPACK and BLAS.

Could you try that out and see if it works? It doesn't need pushing to this PR, but if you can't compile locally then writing code is going to be laborious in the extreme.

I'm currently writing a new Wiki page about installing CVODE and atchem, and that will include a couple of tools to largely automate it, once BLAS and LAPACK are found or installed by the user.

spco:
Great, I was just about to request that final change!

I will squash and merge when this shows as passing.

PR #70, opened 2017-04-12
Move travis checks out of setup_var target and into the main body.

spco:
This stops atchem target being remade when no files have changed.


spco:
I've also removed the redundant dependencies in the listings at the bottom of the Makefile

PR #82, opened 2017-04-14
Small modifications to envVar

rs028:
Correct the equation in calcM to use pressure in mbar instead of torr (I think this is the only place where pressure is used, but need to double check) and rename BOUNDARYLAYERHEIGHT to BLHEIGHT.


spco:
So this latest change makes it fail on Mac (I understand why, and have worked around it on my own Mac, but need to implement a better fix so that Travis can run it successfully).

@rs028 As this is going to be more than one commit, can you move the latest commit to a new branch, to separate it from the other parts of this PR?

spco:
This should be fairly simple:

git checkout master
git checkout -b mod_makefile
git cherry-pick check_envvar

This should create a new branch mod_makefile, and then steal that one commit.  That new branch can then be pushed as a new PR.

spco:
And once that is done, you'll need to re-update this PR by 

git checkout check_envvar
git push --force

which will remove the latest commit here.

PR #84, opened 2017-04-19
Modify Makefile

rs028:


spco:
@rs028 If you're able to update to the latest master branch and check that this works for you still, I'd be grateful!

rs028:
@spco yes it works!

spco:
Fantastic - it took some detective work!

PR #107, opened 2017-04-26
Don't output Jacobian if jacobianOutputStepSize==0

### LINKED ISSUE #35: output of jacobian matrix
The file jacobian.output can be very large (over 1GB). It is needed only for certain types of analysis. The output could be made optionall (eg, if jacobian output step size is set to 0  in modelConfiguration/model.parameters)

spco:
Add outputJacobian boolean, which is set to false when jacobianOutputStepSize is set to 0. This commit breaks travis/tests/short/photoRateCalcParameters.output, where lha and cosx are perturbed. I'm not yet sure why this is.

Closes #35


spco:
The perturbation of cosx and lha is now fixed, by reordering when the jacobian is output to a more sensible location. Also contains some comment fixes, and a tidy up of the test regime to fix the output of instantaneous rates.

spco:
Closes #109

PR #114, opened 2017-04-26
Add numdiff install script

rs028:
This should work. I cannot test the OSX version, though.


spco:
In fact, since FORT_COMP was the only difference between those scripts, we only need one version anyway, just call it install_numdiff.sh.

spco:
Looks good, thanks!

PR #116, opened 2017-04-27
New test scenarios

rs028:
These are 4 additional tests to check that constraining to environment variables and chemical species -- as well as interpolation of the constraints -- work as intended.

At the moment the testing fails, probably because they have been created with the first-commit version.


rs028:
Note: instructions in the wiki regarding `.gitignore` for the test suite are unclear. Should it be in the main directory of the test (`foo`)? what about the constraints directories and `modelConfiguration`?

spco:
I see - can you add another commit with the correct outputs based on the constraints files? Don't worry about copying and renaming by hand - just script it so that all the .output files get copied to the .output.cmp files, etc:
`for file in *.output ; do cp $file $file.cmp ; done`
run in each test folder will copy all those files, and do similarly for instantaneousRates files, e.g.
`for file in *0 ; do cp $file $file.cmp ; done`

Run `make test` again to check that all tests pass before pushing.

spco:
And on the topic of the` .gitignore`, it's probably best to have one per directory, if that directory ever contains things that should be ignored, e.g. the main directory of the test, and the `instantaneousRates` directory, as these have output files we want to ignore. As the constraints directories and `modelConfiguration` only ever have input files, there's not much point using a`.gitignore`, as all those files will always want to be part of the repo. Does that make sense?

spco:
A couple more issues to address that I've spotted so far: `environmentalConstraints` should be `environmentConstraints`. 

Sometimes there are constraint files provided for environment variables that aren't constrained in the `environmentVariables.config` file - is this intentional?

spco:
Let me know when you want me to review this.

rs028:
I am (slowly) working my way through the errors. What is the best way to proceed? put the constarints in the main directory rather than in the travis test directories and rerun them one at the time? 

I also changed the gitignore in travis. I think a centralized version is easier to manage, but let me know if for some reason that's not a good idea.

spco:
The easiest way to proceed is just to run `make test`, as this will run all the tests, and output all the comparison files you need - you just need to rename those files, either by hand or as mentioned above.

I think the `.gitignore` will need work - `.gitignore` is not recursive, so the inclusion of `*.config` currently will only apply to the `tests` directory - you need to put e.g. `!modelConfiguration/*.config`. There are plusses and minuses to having only one `.gitignore` here, but I think it is probably a good idea actually, as each test has identical directory hierarchy.

rs028:
Ok I will fix .gitignore

I am getting this error:

    Running travis/tests/spec_no_env_yes2 ...
    At line 735 of file inputFunctions.f90
    Fortran runtime error: Bad real number in item 1 of list input

But the section of the code it is referring to is new I thinkand I am not sure what it is referring to.

spco:
Which test is this error happening in?

One thing I'd say is to remove the `end` lines in `environmentVariables.config`, as they are no longer needed. Take a look at the existing tests to see that I've removed these lines. This work isn't yet complete though, so some config files still have `end` and some don't. Take a look at an existing test to work out which!

Let me know if that doesn't solve this error.

rs028:
Making progress....

Now I get this:

    At line 760 of file inputFunctions.f90 (unit = 11, file = 'xH���*')
    Fortran runtime error: File   './environmentConstraints/                                                                               BLHEIGHT' does not exist

it makes me think that it is adding extra spaces between the directory name and the file name in inputFunctions.f90
But i should not be the case as there is a trim() command on line 758

spco:
Ok, which test is this updated error occurring in?

rs028:
All of them. Sometimes it gives gibberish:

    (unit = 11, file = 'xH���*') Fortran runtime error: File './environmentConstraints/BLHEIGHT' does not exist

and sometimes nothing:

    (unit = 13, file = '') Fortran runtime error: File './speciesConstraints/CH4' does not exist

The files do exist and the path seems correct to me. My guess is that it is not reading correctly the name of the constraint file, for some reason.

spco:
Hmm, that would seem to be the case. Can you try using `file=trim(fileLocation)` inside both `OPEN` statements?

rs028:
No, it does not work.

spco:
I've just realised that this is because it's looking in the main `./speciesConstraint` directory rather than a user-inputted one (which should be pointing to the test's subdirectory. Pull request wille be incoming in the next 10 minutes, once that is merged then please rebase this branch on top of that, (which will need some merge fixes, as I'll be editing some related lines) and it should hopefully then pass.

spco:
#117 is now in the master branch, so rebasing onto that should be enough. I haven't touched any of the same lines as here, so I'm hopeful the rebase won't even require merges. Let me know how you get on, and whether this fixes things!

rs028:
I am not sure the pro cedure is correct. this is what I did:

    git checkout  master 
    git pull upstream master
    git checkout new_tests
    git checkout mcm_subset.fac
    git rebase master
    make test

without checkout mcm_subset.fac it won't let me rebase. In any case. I get a slightly different error

    At line 760 of file inputFunctions.f90 (unit = 11, file = 'xH���*')
    Fortran runtime error: File 'environmentConstraintsBLHEIGHT' does not exist

There is a `/` missing, but apart from that the filename still appears to be gibberish.

spco:
That looks like exactly the correct procedure.

Once #118 merges (which fixes the slash, but also, importantly, actually makes the tests look in the correct directory) then I hope this will be fixed. Please rebase the same way after 118 is merged.

Sorry about the multiple rebases and slow progress - this is a completely untested part of the code, so I can't tell much of what's going on at your end!

spco:
#118 is now merged.

rs028:
That's all right. Part of the idea of adding  new tests is to check and fix the code, after all!

It's almost there....

    At line 760 of file inputFunctions.f90 (unit = 11, file = 'xH���*')
    Fortran runtime error: File 'travis/tests/spec_yes_env_yes/environmentConstraintsBLHEIGHT' does not exist

spco:
Are you able to push your latest version? I'm losing track of what is and isn't in your version! I'm slightly surprised that the slash has been lost in that example. Does it always crash at this same spot?

rs028:
Damn I made a mistake. I push to "master" not to "new_tests"! How do I go back? :(

spco:
Can you show me the git graph?

`git log --graph --branches --decorate --oneline`

Which commits are on master that should be on new_tests?

spco:
Ah, hang on, you are referring to what's happening on Github rather than locally? If so, make sure you're on new_tests and then do `git push origin new_test` and ensure that shows up here. Then change to master, and do 
`git reset --soft HEAD~1`,
check using `git log` that it all looks like you want it to, and then `git push --force`.

rs028:
Okay, so I wanted to push `new_tests` to this PR, but instead of doing it from the branch, I did from master, by mistake. so now I cannot push new_tests because it is "behind the remote counterpart"

spco:
So, to clarify, you were on `master` and did `git push new_tests` ?

rs028:
No, sorry, I stand corrected. I was on `new_tests` and did `git push origin master`

spco:
Ok, in that case: 
```
git checkout new_tests
git push origin new_tests
git checkout master
git push (--force if it complains)
```

rs028:
That's what I tried, but `git push origin new_tests` gives an error message: " Updates were rejected because the tip of your current branch is behind its remote counterpart"

spco:
Ok, use `git push --force origin new_tests`

rs028:
Okay, sorry for the mess. I need to pay more attention which branch I am in!

spco:
I normally only use `git push`, without specifying the remote or branch, as it defaults to just pushing the current branch to the origin remote. This reduces the possibility of pushing to the wrong place! When rebasing, after `git pull upstream master` I do use `git push origin master`, but even that is probably not necessary.

rs028:
Good idea. I pushed into the PR the current version (error message as in https://github.com/AtChem/AtChem/pull/116#issuecomment-298622052)

spco:
#119 should fix that issue (I've tried it on my machine and it completed, although the outputs were obviously not matching).

spco:
#119 is now merged to master.

rs028:
Yes, finally! I will check the output differences before pushing. Thanks for your patience!

spco:
Based on my limited testing, I think you will need to also remove 'end' from each of `constrainedFixedSpecies.config`, constrainedPhotoRates.config` and `constrainedSpecies.config`, in each test, otherwise you get some odd errors.

spco:
I will do that for the existing tests and base files, in a separate PR.

rs028:
I think only `constrainedPhotoRates`, at the moment.

rs028:
I think I understand all the differences between the new output and the old, except maybe for these files:

mainSolverParameters.output
sparseSolverParameters.output
stepSize.output

Can you check that the differences are due to your changes to the code?

spco:
Hi Roberto, are you able to update your master branch, and then rebase this on top of that? I have some changes I'd like to push to this branch, which fix 3 of the 4 failing tests, but can't overwrite your history in that way unless your history is up to date. Thanks!

spco:
The best way to avoid merge conflicts when you do the rebase on top of master in this context is
```
git checkout new_tests
git rebase -Xours master
```
That will prefer the master branch's changes over yours in the event of a conflict, which is what we need here.

rs028:
I am surprised it failed. `make test` was successfull when I tried before pushing yesterday. But it fails now. I don't get it.
I have rebased as you suggested, but it fails with an error that makes me think it si not reading the filename correctly again

    At line 281 of file inputFunctions.f90 (unit = 11, file = '')
    Fortran runtime error: File 'travis/tests/spec_yes_env_no/environmentConstraints/                                                 end' does not exist

spco:
The issue is that in the meantime I have changed the way that the config files are read, so now the line 'end' is being read as a species to be constrained. That is one of the changes I've made to my version of your branch. Can you push the rebased branch here?

rs028:
All right, but it doesn't let me push it 

git push origin new_tests 
To git@github.com:rs028/AtChem.git
 ! [rejected]        new_tests -> new_tests (non-fast-forward)
error: failed to push some refs to 'git@github.com:rs028/AtChem.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Merge the remote changes (e.g. 'git pull')
hint: before pushing again.

spco:
First ensure your master branch is updated locally with `pull upstream master`, then push that to `origin`. Then switch to `new_tests` branch, and `push --force origin new_tests`.

spco:
Let me know if it won't let you do that.

rs028:
Sorry, I forgot.  I pushed it and the test has run and failed.

spco:
Ok, so I have one remaining issue - we're currently in a scenario where Linux passes but Mac fails. I'm rooting through to find the issue, but I don't think it is directly related to this test - the issue is caused because, when trying to interpolate at interpolationFunctions.f90 line 51, the routine overshoots the final data point provided for PRESS, because the final time of the simulation is 3600 but the final PRESS data point is also 3600. CVODE works by overshooting and then intepolating, but this is breaking the logic of the constraints parser here.

I think the solution here is to give PRESS a value at 3700 rather than 3600. Could you advise whether it's fine to leave the same value, or whether I should use another value?

This only skirts the issue for now, but this PR is already enormous, and this extra bit is not helpful to address here.

I'll open an issue for this, but once you let me know whether to use the same or a different value for PRESS at 3700 (this is in spec_no_env_yes2) I'll merge this PR.

rs028:
@spco yes that is fine.
Actually, I did that on purpose, to check that the interpolation functions were working. If you look at `spec_no_env_yes2.out.cmp` and `spec_yes_env_yes.out.cmp` you see there are a bunch of  `Failed to lin int`. They are not necessarily related to that issue, but I think there is a problem with the interpolation. It may be as simple as a different interpolation method is needed.
I will change PRESS and push in a moment.

spco:
Yes - the issue I currently have is that OSX and Linux seem to output a different number `Failed to lin int` messages, and I'm not certain why - it must be doing a different number of iterations for some reason, but I'd rather merge this PR soon and treat that separately.

See #126 for the issue.

rs028:
@spco Okay. I don't understand what is happening now. My master branch is even (both local and on github). My new_tests branch already has PRESS timestamp at 3700. So I have no changes to make, but the tests still fail on my machine. Am I missing something?

spco:
Is it just the one test `spec_no_env_yes2` that's failing?

rs028:
no. it's all of them, with the same error as this morning "Fortran runtime error: File 'travis/tests/spec_yes_env_no/environmentConstraints/                                                end' does not exist"
It looks like I am behind, but it tells me I am even. Also how did the change to 3700 ended up here?

spco:
show me `git log --branches --graph --oneline --decorate`

rs028:
* d5eb9ca (HEAD, origin/new_tests, new_tests) Correct reference outputs for new tests
* dce6b0b Remove whitespace from fileLocationPrefix - 2
* ab0579f Remove whitespace from fileLocationPrefix
* 03a3284 fac file needs to be as downloaded from MCM website
* 59cc1b9 Remove 'end' from config files
* 7e79fb9 Correct variable names
* 74c2d81 Rename environmentConstraints dir and delete unnecessary constaint files
* 756e024 Simplify .gitignore for travis directory
* 456b2da Add the constraints files
* 2cdff9a Add new tests to Makefile
* 3e82a58 Add 4 tests to check constraints and interpolation
* e1cc1bf (origin/master, origin/HEAD, master) Remove 'end' from constrainedFixedSpecies.config, const
* 3a9d43c Correct the concatenation of slash to constraint filepaths.
* 2328246 Fix slash, and call cmd args in test_runner.sh.
* 3b67a3e Add parameters env_constraints_dir and spec_constraints_dir to module directories, and use.
* 8253d04 Add numdiff install script (#114)
| * 7ea60f0 (origin/install_numdiff, install_numdiff) Remove previous scripts
| * 81958ec Merge linux/macos scripts
| * 14f03be Version for osx; add some comments
| * e79bb81 Add numdiff install script
|/  
* 0c4f285 Remove unnecessary explicit import.
* 95cf4c4 Use envVarNames and numEnvVars directly from module in getEnvVarNum, rather than passing in 
* 8497601 Rearrange order of arguments of getEnvVarNum() to put output last.
* c04e9dc Set intents within getEnvVarNum(), and comment.

spco:
Ah, we've lost the commits I put in. My fault. I'll add those on top.

spco:
But first, can you push your latest commit, with the edited PRESS value?

spco:
Just `spec_no_env_yes2` PRESS value is currently at 3600, and needs to move to 3700 or something.

rs028:
According to my system everything has been updated and there is nothing new to push.

spco:
Hm, ok. Leave it with me.

spco:
Hmm, I currently have the issue that the OSX and Linux versions give different results, just for `spec_no_env_yes2`. I think there is a sensitivity somewhere, which explodes due to rounding errors. I haven't yet got to the bottom of what specifically is causing it, or how to fix it.

What I propose is to extract that test out of this PR and into a new one, for further testing later. This PR can then just contain the 3 tests, and should pass.

Does this sound reasonable? I'll take the lead, as it will require some git-magic (and trial-and-error ;) ) to extract the necessary bits.

spco:
I've gone ahead and split the failing test off from the other three - that means we can incorporate the changes here into other PRs, as that's holding things up a bit. I'll open a PR afterwards which contains the 4th test.

rs028:
Sorry, I thought I had replied, but clearly my message got lost. I think it is a good idea. I am also getting a little bit worried  about these interpolation / rounding errors.

spco:
@rs028 Never mind, I guessed as much. Thanks for bearing with us on this PR. It's been pretty epic, but exposed a number of flaws, which is what this all about!

The interpolation and rounding errors we can definitely look into. Perhaps we need another test or two, less complicated than `spec_no_env_yes2`, to delve deeper and work out what's going on. My current work on simplifying and unifying the code should help with that, so I propose to park this for a little while, and come back to it when the code is in better shape?

rs028:
@spco All right. Yes it is good that we catch all these problems.
I will go ahead with my original task of sorting the DEC, SZA and H2O calculations.

PR #118, opened 2017-05-02
Fix slash, and call cmd args in test_runner.sh.

spco:
Fixes the incomplete #117 implementation.


spco:
I'm skipping the OSX testing, as the Linux passes and I want to get this moving.

PR #123, opened 2017-05-02
Make printing of excerpts of lists more standard

spco:
This should really be handled by a generic function...


spco:
Generic function left for another day.

PR #128, opened 2017-05-04
Add spec_no_env_yes2 test files, and add to Makefile.

spco:
Based on @rs028 's work, originally included in #116, but currently doesn't pass Travis tests because of difference between OSX and Linux output. Not sure whether this is just numerical instability due to an unhelpful setup, or whether it's a bug with 2 environment variables being simultaneously constrained.


rs028:
**Note**: the wiki page on adding tests will need updating and clarification after the changes in #116 (.gitignore, constraints directories, etc..)

spco:
This is now merged, but issues still remain - I can only keep this test stable if I reduce the time of the simulation to about 300 seconds. This is annoying, but we'll have to live with it for now until we better understand the lack of stability.Either way, this test should pick up if we break something, which is the main point.

PR #131, opened 2017-05-05
Integer kind

spco:
Massive change - most things end up inside modules, integer kind is added.


spco:
Superceded by #132, #134, #136.

PR #132, opened 2017-05-05
Make modular and introduce species integer kind NPI

spco:
Make longint_kind consistently take argument 10.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ makefile.local:
 
 include makefile.local
 
-SRCS = dataStructures.f90 configFunctions.f90 inputFunctions.f90 atchem.f90 mechanism-rates.f90 solverFunctions.f90 outputFunctions.f90 interpolationFunctions.f90 constraintFunctions.f90 instantaneousRatesFunctions.f90 conversionFunctions.f90 utilityFunctions.f90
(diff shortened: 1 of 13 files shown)


spco:
@rs028, I'm making almost all the files into modules, but at the same time I'm neatening up the use of integer and real kinds to ensure portability. Specifically, gfortran differs in its handling of quad-precision real numbers. Using kinds from `types_mod` makes it portable. I'm hoping this may hold the key to why test output for `env_no_spec_yes2` differs on Linux and OSX. We'll see.

One question I have now is whether we want to drop the maximum integer size for numbers of species. In this PR, I've introduced kind=NPI for all variables handling numbers of species. Judging by the test results, setting this to INT64 gives me identical results, while INT32 gives me differing results. I have yet to fully check the diffs, but I think this is just a difference in the number of iterations the solver takes etc. Since INT32 is generally cheaper than INT64, I would suggest we use that. INT32 still has a maximum value of ~4.3bn, so is way above what we require for the number of species or the number of reactions.

rs028:
@spco sounds good to me. Feel free to move around subroutines so that they end up in the module where it makes more sense.
As for the other questions, I am not sure I understand, or maybe we are reaching my fortran knowledge limits :) As long as we don't sacrifice accuracy in the results, I think I will trust your judgement on this.

spco:
@rs028, OK, I think I will change to INT32. Essentially we are reducing the maximum number of species we can hold from 1.8e+19 to a mere 4.3e+9, which I think is acceptable :) In return, we halve the memory used to store each of those integers. 

I'll also look into moving subroutines as you say - there are a few that probably do need moving, or just a few modules that could be clumped together.

PR #135, opened 2017-05-08
H2o conv

rs028:
calcM is okay (I just changed the order of arguments to have a consistent style: output, inputs)
convertRHtoConcH2O becomes convertRHtoH2O and for the moment only adds pressure as an unused argument. The conversion equation is not yet changed.


spco:
Could you change the subroutines convertRHtoConcH2O() and calcM() to be functions? By the look of things they should have no side-effects, and output a single variable, so they would be better as functions. Could you also add `intent(in)` to the input arguments? See e.g. `matchOneNameToNumber()` definition for the syntax. This ensures you can't misuse an input argument inadvertently.

For future reference, I'm working to the style of variables being in the order (inputs, outputs), although in the case of functions that doesn't occur.

spco:
Whenever a variable is referenced within a subroutine/function, you have to use the name declared in the subroutine header, not the name used in the main routine. If you use the same name as in the main routine, then that's ok but a bit confusing - better to use a different name inside the subroutine, as then it's obvious which you are referring to. Subroutines/functions cannot access variables that aren't passed in as arguments, except in the case where we use modules to pass around variables.

So essentially you can always use a subroutine, but a function can be used in special circumstances to 'lock down' usage, to avoid unexpected side-effects.

The difference between a function and a subroutine is quite simple: a function can take N inputs, but output only 1 output. It is not allowed to change any of the input variables.
A subroutine can take as many arguments as you like for input and/or output. They are allowed to change those variables in any way you like. The use of `intent(in)`, `intent(out)` and `intent(inout)` allows you to outlaw certain uses - a variable declared as `intent(in)` cannot have its value modified inside the subroutine, while one declared `intent(out)` cannot have its value used until it's set inside the subroutine. `intent(inout)` allows both uses.

Finally, when calling a function, you need to use e.g. `M = calcM (pressure, temp)` rather than `CALL calcM (M, pressure, temp)`.

rs028:
All right. I agree that the name of the variables with functions/subroutines should be different than from the main program, even if just for clarity. But then there are quite a few variable names to be changed here!

spco:
There's a typo in line 55 which is stopping this from building - it's worth trying to build locally before pushing, as it's much faster and easier to debug there rather than via Travis.

rs028:
That's what it was! The RH to H2O conversion is not finished but for the time being it can stay as is (if you need to merge, otherwise leave this PR open) . I need to do a few more tests to be sure of the equations.

spco:
No pressure from me to merge this, it's not holding me up.

rs028:
@spco I see you have changed quite a few things :) 
I am trying to catch up. just be sure:

    IMPLICIT NONE
    DOUBLE PRECISION 

changed to this? both in subroutines and functions?

    USE types_mod
    real(kind=DP)

spco:
I'm working on a bit of styling, and will put up a style guide soon, which should explain things! And yes, there have been some large-scale changes to the code structure, so I'd recommend making sure your local git repository is updated, and branches are rebased.

`types_mod` is a module in `dataStructures.f90`, which defines what data types there are in a portable way - the problem is that different compilers and architectures may define `DOUBLE PRECISION` in a different way - for example `gfortran` may allocate 2 bytes, while `ifort` could allocate 4 bytes. This causes havoc with reproducing results on different machines, so rather than rely on `DOUBLE PRECISION` we use `DP` as defined in `types_mod`, which is entirely portable - all compilers must handle that the same way in order to comply with the Fortran standards.

So `DOUBLE PRECISION` gets replaced by `real(kind=DP)` everywhere. `USE types_mod` lets the subroutine understands what `DP` means: without this, the subroutine doesn't know.

The same is true for integers, in a similar way. I'm using `NPI` to denote the type of integer we want to use when we are counting species, or environment variables.

`IMPLICIT NONE` is not related to that change. All subroutines should contain `IMPLICIT NONE` (just after the `USE` statements, and before the variable declarations). This tells the compiler to demand that all variables are explicitly given a type - one of the oddities of older Fortran is that variables were given a type implicitly if they weren't defined, based on the first letter of the variable name - a bizarre idea, and a minefield for bugs, but there we go. `IMPLICIT NONE` ensures we can't accidentally have implicitly defined variables.

rs028:
Yeah, that's what I am doing, rebasing :)
Okay, I understand.  I have seen  `USE types_mod` used not in every single subroutine but at the beginning of the module. Is that better?
Also `IMPLICIT NO` is required for functions as well?

spco:
Yes, in a couple of situations if I know pretty much every subroutine in a module requires `types_mod` then I've stuck it at the top of the file to keep it tidy. Either is acceptable and has up or downsides.

Yes, `IMPLICIT NONE` is required for the main `PROGRAM ATCHEM` and all subroutines and functions, as well as modules (there may be some missing, feel free to add if you find any).

rs028:
okay, one more question. I have been thinking about what you said about variables having different names in subroutines/functions than in the main code. Is it worth for me to change the names in the subroutines I modify or  will I mess up unnecessarily?
If it is okay should we agree on a common nomenclature? eg, short names for subroutines (press) and long for main code (pressure)? or whatever you think makes more sense.

spco:
We can agree a nomenclature if that helps, but I think it's probably difficult to come up with one that fits all scenarios. Some functions are very generic, and so should have very generic names, but others are more specific and are only used once.

One thing I would suggest is that, in general, the subroutines should use more 'generic' names, rather than the specific name of what will be passed. Rather than naming the argument after what will be passed in in this specific case, try to name it after the role of that argument _within the function_, rather than keeping the meaning of where it is used within the main function.

I'm happy to bounce ideas around if you have a good nomenclature system - I'm wary of just shortening names in all circumstances.

rs028:
Okay, I have added one function, but also done some cosmetic changes and added indentation and a separator between subroutine in two files. See how you like it (I hope I did not break anything!)

spco:
I'll wait to see whether the test passes. Annoyingly, diff thinks lines with whitespace changes are new, so the diff I can see is rather unhelpful for review! Never mind.

Do you have a particular purpose for the `calcPRESSfromM` function? I'd prefer if it were renamed `calcPressureFromM`, I think? `PRESS` is our name for a variable, but `pressure` is the name of the actual scientific quantity.

rs028:
I think there is a way to tell diff to ignore whitespace changes but I don't know if it is possibile to do that on github.

I have changed the name of the function, as suggested. It makes sense to call the variable with the full scientific name in the main code and with a shorthand or generic name in the functions/subroutines.

The reson for the new function (now called simply `calcPressure`) is that M and Pressure can be calculated from each other. It would be wise to stop the user from inputting both of them (either as set values or constraints) because it could introduce inconsistencies in the model. So I think that if M is constrained or constant then Pressure must be calculated and viceversa. But neither function can be NOTUSED and they cannot be both calculated or both constrained/constant. I will change the wiki page accordingly, although it is not implemented yet. Does it make sense?

spco:
Yes, that makes sense - inconsistent input data is a nightmare! Make an issue to keep track of the implementation of that feature? It might take a couple of PRs and wouldn't want to be missed.

One question I have is how does this depend upon the temperature? Is the temperature always given, or is that calculated from something else? I'm just wanting to check that we can't end up with an indirect dependence upon M or pressure again!

rs028:
Temperature should also be given. I am not sure of the purpose of the subroutine in `utilityFunctions` so I may end up deleting it altogether, but first I want to understand what it does :)

If this branch tests okay, could you merge it and close it please? I won't be able to work on the code for the next wee, and I don't want this branch to fall so behind the master that it becomes a nightmare to rebase it.

spco:
Ok, looks good - I'll squash and merge it now.

Yes, `temperature()` in `utilityFunctions.f90` does seem to be unused. I wonder whether it's supposed to be used near where `atmosphere` is used in `mechanism_rates()`. Either way, it currently does a bit of nonsense anyway, so would need fixing if we do use it.

PR #140, opened 2017-05-09
Solver funcs

spco:


spco:
This was superceded at some point.

Issue #158, opened 2017-05-17
Swap argument order of calcDec


spco:
To fit input-output order.

spco:
Closed by #159

PR #159, opened 2017-05-17
Make calcDec() a function.

spco:
Add intents to all functions in conversionFunctions.f90, and change calcDec() to a function.


spco:
Closes #158

Issue #63, opened 2017-04-07
Remaining files in modelConfiguration should be input dependent


spco:

spco:
`mechanism-rate-coefficients.f90` and `mechanism-rate-declarations.f90` can't be moved, since the compiler needs to know their position at compile-time, but the path to `modelConfigurations` subdirectory is only given at runtime. Trying to enforce that those are both the same is asking for trouble. So we'll just place these in the main directory (not `./modelConfiguration/`), but move `mechanism.species/reac/prod` to the runtime-defined subdirectory.

spco:
Closed by #161

PR #161, opened 2017-05-18
Move mechanism file defaults, allow direction to alternative directories

spco:
Add a second argument to `tools/build.sh`, `tools/setup_atchem.py`, and `tools/mech_converter.py:convert()`, denoting where the `mechanism-rate-coefficients/declarations.f90` files should go when built (which now defaults to '.' , and there's no real way to pass this on to the `Makefile`, so the `Makefile` will break if you send this anywhere else).


spco:
Fixes #63

PR #163, opened 2017-05-18
Make variable constrained species data input more robust

spco:
Count lines in data files of constrained species, rather than rely on a number on the first line, which has to match the number of remaining lines in the file.


spco:
Also test for existence of all files from inputFunctions.f90, via `inquire_or_abort()`.

Issue #4, opened 2017-03-17
Fix the RO2 sum function


spco:
It currently just outputs zero, but really it should be outputting the sum of the concentrations of all RO2 species at the current timestep.

rs028:
Need to check that the variable has a value before being set to zero for output to file. It is potentially a serious bug.

spco:
The code in mechanism-rate-cofficients.f90, which calculates the ro2 sum, is only INCLUDEd in mechanism_rates() in mechanism_rates.f90.
mechanism_rates() is called by resid() and jfy().
jfy() is called each time Jacobian output is requested.
resid() is called by FCVFUN(), which is called by the user-defined FCVJTIMES().
The user-defined FCVJTIMES() is only called when fcvspilssetjac() is called with non-zero flag, which only happens if JvApprox==1.

So, the situation we currently have is that ro2 will only be calculated when JvApprox==1, otherwise it is not calculated. This is independent of the call to ro2sum() and outputEnvVar() in the main loop. So if this flag is not set, then nothing will be calculated for ro2.

spco:
Correction: FCVFUN() is called independently of FCVJTIMES() (which in fact is never called), and so ro2sum() will always be called.

spco:
Fixed by #183

rs028:
@spco Just one question. Was this only a problem with the output of the variable or was the variable also incorrectly calculated?

spco:
It was only an issue with the output, although it could easily have affected the calculations - luckily it didn't, due to the places it was used.

rs028:
That's a huge relief! :)

PR #183, opened 2017-05-24
Fix ro2sum()

spco:
Split the generated list of ro2 sum into a separate file, ro2_rates.f90. This is generated by tools/mech_converter.py. This is then included inside the ro2sum() function (now a function not a subroutine), and ro2sum() is called within mechanism-rate-coefficients.f90, as well as elsewhere.


spco:
Fixes #4

Issue #156, opened 2017-05-16
cubic spline interpolation (ln)


spco:
Choosing `interpMethod ==2` in `getConstrainedQuantAtT2D()` leads to exponentiating the concentration. Is this ever un-exponentiated anywhere else?

spco:
Closed by #188

Issue #186, opened 2017-05-25
Cubic spline second derivatives


spco:
@rs028 : `getConstrainedQuantAtT()` uses the function `cubic_spline()` to calculate the cubic spline interpolating the given data points, to try to give a smooth representation of the constraint data, which is helpful. However, calculating the cubic spline requires not just the value of the data at each time point, but also the second derivative at each time point.

The current code is semi-written to make use of these second derivatives - there's some code commented out for reading the data in, and then both `getConstrainedQuantAtT()` and `cubic_spline()` use this data. However, since the read-in code is commented out, the data for the second derivatives is either zero or (perhaps worryingly) not set before use. My question is this - are any constraint data sets going to have this second-derivative information attached, as an extra column in the input file? If not, then the extra code should be removed. If this _data_ may be provided, then we need to properly handle the case where it is not provided.

Thoughts?

rs028:
Hi @spco I am back. I had a good chat about what we are doing with the Leeds/York people and they are all enthusiastic. 
Re this question, I don't think the second derivative is likely to be attached to the constraint data. My understanding (from `modelConfiguration/model.parameters`) is that there are several interpolation methods available. I think the important thing is that they work as intended and the implementation is correct. Does it make sense?

spco:
Hi @rs028 , that's great. Would you be able to meet up and chat through some things (both low-level questions as well as discussing the progress of the project)?

I guess on the interpolation front, what do we want? If the user asks for a cubic spline interpolant, but doesn't give any 2nd derivative data, or submits that data to be 0 everywhere, then you end up with exactly the linear interpolant, which is implemented as a separate option.

So, if second-derivative data is never going to be supplied, we have two options: (1) remove the cubic-spline interpolation as an option; or (2) implement a _different_ cubic spline interpolant, called the 'natural cubic spline', which doesn't require this second derivative information, but *does* require the solution of a linear system of equations of the size of the number of data points for that dataset. This number isn't large, but this would require some extra machinery to implement.

My suspicion is that this option has never worked properly, and has always just output the linear interpolant anyway. This aligns with the issue highlighted in #156, which I don't think has ever worked either.

My suggestion is that we remove both the `cubic spline` and cubic spline (ln)` options (options 1 and 2) from the interpolation options. It has probably never worked, it would take a fair bit of work to make it work, and it isn't my priority right now. If they are wanted in the future, then the code will still be available in the history of the GitHub repository, but it would be better to clean it out of the release version now.

[...]

PR #188, opened 2017-05-25
Interpolation methods

spco:
Remove cubic spline interpolants (methods 1 and 2) as they've never worked. Renumber 3 (left-handed p.w. constant interpolation) and 4 (p.w. linear interpolant) to 1 and 2 respectively. Fix all test input files.


spco:
Closes #156 and #186

Issue #133, opened 2017-05-07
check M calculation


rs028:
Need to verify the output  of subroutine calcM (related to #23)

spco:
Can I assign this to you? It would be helpful if you link to the source of the calculation (a paper etc) when you get round to it.

rs028:
Yes. I will put in a PR for both #133 and #23

rs028:
This subroutine is correct. I will just clean up and comment the code.

![m](https://cloud.githubusercontent.com/assets/7703778/25801973/1cd05ae0-33e8-11e7-8419-8ce1efc8b499.png)

rs028:
Make sure that input of M and Pressure are consistent - related to #23

rs028:
@spco Good stuff! :+1:

PR #194, opened 2017-05-26
Press m

rs028:
This is meant to solve issue #133 (plus some tidying up and renaming of a couple of functions).
The main objective is to prevent the user from inputting both M and PRESS.
If M is fixed value or constrained, then pressure should be calculated.
If M is calculated, then pressure should be  fixed value or constrained.

Neither can be NOTUSED, but they cannot be both calculated or both constrained, to ensure there are no inconsistencies between two closely related variables. See wiki page: https://github.com/AtChem/AtChem/wiki/Environment-Variables


spco:
I think the logic of this doesn't quite work as it stands. The loop starting at line 176 loops over each of the elements of `orderedEnvVarNames`. This defined order means that pressure is _always_ set before m. So `m_set` is never going to be `.true.` when the pressure is processed. But equally, if we swapped the defined order, then when we get to m, pressure will not have been processed, and so `pressure_set` will never be `.true.` there. We need a change to the mechanism to allow an either/or choice, if that makes sense?

While you're there, I'm not sure how old the base version of the `constraintFunctions.f90` is you're using? I recently moved the `use types_mod` line out of the module heading, and placed it into each subroutine as necessary, as that's probably in better keeping with the style elsewhere.

rs028:
I just realized that myself. Possibily the easiest solution is to get rid of M as an input altogether, just input Pressure and always calculate M.  What do you think?

spco:
To clarify/refine what I said a moment ago, I actually think the mechanism of the `m_set` is ok, it's just that, because we have a defined order, we can never calculate pressure, because it's never called after calculating m.

I think we'd need to put a piece of logic in _before_ filling `orderedEnvVarNames`, which decided which of m or pressure should be done first: if m is calculated and pressure is not, then 
```
orderedEnvVarNames(1) = 'PRESS'
orderedEnvVarNames(2) = 'M'
```
If press is calculated but m is not, then 
```
orderedEnvVarNames(1) = 'M'
orderedEnvVarNames(2) = 'PRESS'
```
If both are calculated or both constrained, then error out.

Then, carry on filling `orderedEnvVarNames` and continue with the code as you've added it in.

rs028:
As for types_mod I though better practice was to put it at the beginning and not in each subroutine, so I have been moving all of those, but I'll change it back as it was.

spco:
On the types_mod: sorry for the confusion, I think my mind has changed on which is better practise since I last spoke about it!

On the issue of PRESS and M, I'm not qualified to make that decision - is pressure always available with the data? If so, then sure let's do that.

rs028:
No problem.
Yes Pressure is most likely available. Or rather you can't measure M, so you must always calculate it from PRESS and TEMP. I think it makes more sense to get rid of it. Will make life easier.

spco:
OK. If so, need to decide how to handle the input of M: demand it be in `environmentVariables.config`, but always as 'CALC', or remove it from the file? If the former then it's a bit odd only allowing one possible option, but if the latter then it still needs placing within `envVarNames` and `envVarTypesNum` to allow it to be handled easily in e.g. `getEnvVarsAtT()`.

rs028:
Yes, I think you are right. If I remove the string from `environmentVariables.config` is better. Then orderVarNames does not have to change. The code will just have one option , that is to calculate M on line 189. Is that right?

spco:
I think so. Just need to be careful that M is still treated like an environment variable (it needs to be inside `envVarNames`, for example, in order to be outputted in `outputEnvVar()`).

rs028:
I think it would simplify a lot the user interface, but I am starting to regret going into this :(
If M disappears `environmentVariables.config` then envVarNuyms is no longer correct and M is not in the list of `envVarNames`.

As a provisional solution I tried to leave it in but set to CALC, just to see if it would work and it runs on my machine (the tests fail, but I think that's because  `environmentVariables.config` has not been changed yet for those; I'll fix it in the coming days).

spco:
Thanks for this @rs028 . Looks good. I'd agree that it's best at this point to just leave `M` as `CALC` everywhere, and come back to this in a separate PR if we want to treat it differently.

I'd be eager to split this up as it stands a little anyway - the changes to `atmosphere()` for example are not related to the rest of this PR, and will probably cause conflicts with half of what is done in #198. Otherwise, this looks good to go - you're right that the test failure is just a case of changing the default value in the comparison output. Will you be able to get round to this either today or Tuesday? If not, I'm happy to take it from here and get it merged, so some of the others can also get close to being merged.

spco:
I've just resolved the conflicts here brought in by merging #197 a moment ago.

spco:
Once the remaining reference to `calcM` is updated, would this be good to go from your perspective? If so, I'd be keen to see this merged next.

rs028:
Yes, the current push is just to sync my branch. I need to fix the tests, I should have time to do it today.

spco:
Unfortunately I've been very busy today in changing the screen output, and have merged a few bits, hence the conflicts - sorry! I'll sort the pain out. From your side, are you happy for this to be merged?

rs028:
I think it's my fault, as I was running too many things in parallell. Anyways, yes, if it is easy for you to sort this, then it can be merged.

spco:
Not to worry - this is always the messy part!

spco:
Right, hopefully that passes...

Issue #23, opened 2017-03-21
RH and H2O


rs028:
Is RH calculated correctly from H2O (and viceversa) when one is set to a constant value or constrained and the other is not?

spco:
@rs028 , can you point me to what the correct calculation is to convert between RH and H2O?

rs028:
@spco I will check and fix the equations of this and #14. The various options for the switches of the envVar are in the wiki (see issue #22)

spco:
@rs028 Ok, great. I'm thinking of working on #22 shortly, to put in the correct defaults as per the wiki.

rs028:
@spco Okay. See also #83 (it may be just a cosmetic issue, but it may be important, I am not sure) .

rs028:
The current equation to convert RH to H2O gives correct results but underestimates H2O at high relative humidity (see plot below). This is possibily due to lack of pressure dependence.

![rh_h2o](https://cloud.githubusercontent.com/assets/7703778/25781287/dde0e3bc-332f-11e7-84e0-d7e9dd6c57ad.png)

spco:
Can you link to the source of the current calculation, and to any more accurate formulae?

rs028:
I have done the comparison a little bit better. I am not sure where the current implementation is taken from.
I have used the _Humidity Conversion Formulas_ document on the Vaisala website to convert the relative humidity into volume of water vapour per volume of dry air (PPM), then multiply the result by 1.0e-06*M to get the molecule cm-3.

PR #197, opened 2017-05-27
Fix for issue #23

rs028:
Add pressure dependence to RH-H2O conversion and correct the equations. Also fixes a mistake in calcM (to be renamed calcAirDensity after PR #194). It also removes calcPressure, as it will no longer be needed after PR #194.

### DIFF
--- a/conversionFunctions.f90
+++ b/conversionFunctions.f90
@@ -49,7 +49,7 @@ contains
     temp_c = temp - 273.15
     wvp = (rh/100) * 6.116441 * 10**( (7.591386 * temp_c) / (temp_c + 240.7263) )
     h2o_ppm = 1.0d+06 * wvp / (press - wvp)
-    h2o = h2o_ppm * calcM(press,temp) * 1.0d-06
+    h2o = h2o_ppm * calcM(press, temp) * 1.0d-06
 
     return
   end function convertRHtoH2O
--- a/conversionFunctions.f90
+++ b/conversionFunctions.f90
@@ -38,22 +38,6 @@ contains
 
   ! ----------------------------------------------------------------- !
 
-  pure function calcPressure( m, temp ) result ( press )
-    ! calculate pressure (mbar) from the number density of air
-    ! (molecule cm-3) and temperature (K)
-    implicit none
-
-    real(kind=DP), intent(in) :: m, temp
-    real(kind=DP) :: press
-    real(kind=DP) :: press_pa
-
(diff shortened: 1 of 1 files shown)


rs028:
Note: only the style test fails.

conversionFunctions.f90
```diff
@@ -31,42 +31,26 @@ pure function calcM( press, temp ) result ( m )
...
+    ! pressure in mbar (1 mbar = 1 hPa), temperature in K
     implicit none
 
     real(kind=DP), intent(in) :: rh, temp, press
-    real(kind=DP) :: h2o, exponent, e1
+    real(kind=DP) :: h2o, h2o_ppm, temp_c, wvp
 
-    exponent = exp( -1.00d00 * ( 597.30d00 - 0.57d00 * ( temp - 273.16d00 ) ) * &
-                   18.00d00 / 1.986d00 * ( 1.00d00 / temp - 1.00d00 / 273.16d00 ) )
+    temp_c = temp - 273.15
```
spco: Could you write a comment detailing where this calculation comes from, both the document and the equations? As far as I can tell, it looks like the combination of (6) and (1), then (18) in the Vaisala document. The conversion from h2o_ppm to h2o I also understand, but is there an explicit formula we can link to, or at least write in the comment?

conversionFunctions.f90
```diff
@@ -31,42 +31,26 @@ pure function calcM( press, temp ) result ( m )
     real(kind=DP), intent(in) :: press, temp
     real(kind=DP) :: m, press_pa
 
-    press_pa = press * 1.0d+2
-    m = ( 6.02214129d+23 / 8.3144621 ) * ( press_pa / temp )
+    press_pa = press * 1.0d+02
+    m = 1.0d-06 * ( 6.02214129d+23 / 8.3144621 ) * ( press_pa / temp )
```
spco: You previously said that this calculation was correct: https://github.com/AtChem/AtChem/issues/133#issuecomment-299840245 . Has this now changed, or is there confusion over the multiplications by 1e6 and 1e-6 at lines 51-2?

rs028: Okay, now I am getting confused. I will double check later today or tomorrow.

rs028:
If this passes the tests, lets merge it. I will address the issue with the equation for M in PR #194.

rs028:
Ok, good to go.

rs028:
Yes, the multiplication for 1e-6 should be in there. I think what happened was an error in copy&paste, because I tested the function externally. PR #194 is ready anyways, just need to fix the test results.

Issue #108, opened 2017-04-26
zenith() and ramp() need calculations checking and tidying.


spco:
The workings of zenith() and ramp() are very flaky. I'm not entirely sure of the calculation that _should_ be going on - @rs028 are you able to point me to a standard work, or other place, where this calculation is defined?

Issues I've spotted:

- [ ] What is ramp() calculating? It seems to output max{0,x} - is that what's required?
- [ ] zenith() has an input argument theta which is calculated but never used.
- [ ] The secx calculation seems like a fudge, adding 1e-30 to avoid a division by zero - what is the real calculation supposed to be?

Related to #14 ?

rs028:
@spco I am not 100% sure, but I think ramp() is basically a way to deal with negative numbers.  Like the usage of 1e-30 (which may appear in other parts of the code) I think these are leftovers from the model written for the FACSIMILE solver.

zenith() is related to #14. I will look into that.

rs028:
This can be closed. The remaining issues are related to issue #14 and are being addressed in PR 198.

spco:
Ok. I may open another issue to look at the 1e-30 issue (because it would be great if we handle the case of 0 properly, rather than fudging it), although that also comes under the JFac umbrella as well.

Issue #14, opened 2017-03-21
calculation of DEC is likely wrong. Check that SZA is correct.


rs028:
The calculation of the sun declination (DEC) is likely wrong. It should be:
- at summer solstice = +0.41 rad
- winter solstice = -0.41 rad
- spring and autumn equinoxes = 0 rad

It may be related to longitude east being positive rather then negative. This needs to be tested on a range of dates and latitudes/longitudes.

rs028:
Sun declination is not related to latitude/longitude but is a function of time only. The equation in calcDec (constraintFunctions.f90) is off by about 11 days. Need to fix this and check that the SZA is calculated correctly (in zenith.f90)

spco:
I'm currently working on a PR to fix the calculation of dec, using the equation from https://en.wikipedia.org/wiki/Position_of_the_Sun#Declination_of_the_Sun_as_seen_from_Earth

rs028:
I was going to do that, but I haven't started yet, so if you prefer to do it yourself, I can focus on the test scenarios.

PS: there a couple different ways to do that. The function fSolar in https://github.com/rs028/atmosch-R/blob/master/atmosPhys.R also does that with a different approach. In principle they should give the same (or at least very similar) numbers, so it can be used to double check the results. Sound reasonable?

spco:
#112 fixes the calculation of DEC with a more accurate version. SZA has not been checked.

rs028:
@spco Can you assign this to me?

rs028:
@spco, I am finally getting to the bottom of this. 
Both the old and the new equations for DEC are wrong (I think the one you used is for astronomy, which is a slightly different calculation). Should I wait until we have merged PR #206 and #208 or can I just branch off the present master? I don't want to do a messy reconciliation.

spco:
Feel free to make a change to the DEC calculation straight off the latest master - perhaps a better version of that will solve my numerical issues! I can handle the messy reconciliation later, just keep the PR very focused on that one change (plus resultant changes to output files).

rs028:
Okay. do you want to merge #206 now? It is not complete, but it is all set up to make it easy to add comments and descriptions. Since it is a lot of files, may be easier to it now.

spco:
[...]

PR #198, opened 2017-05-29
Sun Angles 

rs028:
PR to address issues #14  and #108. The equation for SZA is correct, but need to check the calculation of LHA and DEC.

Question 1: should variables `secx`, `cosx`, `lat `, `longt`, `lha`, `sinld`, `cosld` be arguments of the subroutine `zenith` now `calcZenith`? These are written to `photoRateCalcParameters.output` for diagnostic purposes

Question 2: are `latitude` and `longitude` global variables?


rs028:
Wait to merge untile LHA and DEC calculations have been checked

utilityFunctions.f90
```diff
@@ -28,35 +32,75 @@ subroutine atmosphere( m, o2, n2 )
...
+    ! calculate cosine of solar zenith angle (cosx)
+    ! SZA is the angle between local vertical and center of the Sun
     sinld = sin( lat ) * sin( dec )
     cosld = cos( lat ) * cos( dec )
+    cosx = cos( lha ) * cosld + sinld
 
-    rampValue = ramp( cos( lha ) * cosld + sinld )
+    ! set negative cosx to zero and calculate the inverse (secx)
+    ! secx is needed for the MCM photolysis parameterisation
+    rampValue = ramp( cosx )
```
spco: Could we perhaps rename `rampValue` to `ramped_cosx` or similar, to be more descriptive of what that value is?

utilityFunctions.f90
```diff
@@ -28,35 +32,75 @@ subroutine atmosphere( m, o2, n2 )
...
+    maxDecInRad = 23.44 * pi / 180.0
+    currentFYear = dayAsFractionOfYear + ( t / secondsInYear )
+    temporary = 2.0 * 0.0167 * sin( 2 * pi * ( currentFYear - ( 2.0 / daysInYear ) ) )
+    dec = asin( sin( -maxDecInRad ) * cos( ( 2.0 * pi ) * ( currentFYear + ( 10.0 / daysInYear ) ) + temporary ) )
+    return
+  end function calcDec  ! NB: is this in radians or degrees?
+
+  ! ----------------------------------------------------------------- !
+
+  subroutine calcZenith( ttime, dec )
```
spco: Can we rename `ttime` to `t`, which fits the style of other functions?

utilityFunctions.f90
```diff
@@ -28,35 +32,75 @@ subroutine atmosphere( m, o2, n2 )
...
-    cosx = cos( lha ) * cosld + sinld
-
     return
-  end subroutine zenith
+  end subroutine calcZenith
+
+  ! ----------------------------------------------------------------- !
 
   pure function ramp( arg1 ) result ( rampValue )
+    ! if arg1 is positive returns arg1
```
spco: While you're here, could you rewrite this function in a more transparent way? I'd prefer to see the function use `rampValue = max( 0, arg1 )` than the current formula (it gives the same result, but is much easier to understand!)

utilityFunctions.f90
rs028: Actually, do we really need this as a separate function? I believe it is used only here. What if I just substitute lines 93-94 with:

    ramped_cosx = max(0, cosx )
    secx = 1.0 / ( rampValue + 1.00d-30 )

spco: That sounds a good idea!

rs028:
Okay, all done. What about my questions above?

spco:
Ah, I forgot to answer those. (1) I think it's best to leave these as members of `zenithData`, rather than as parameters to pass in and out - it's easier and tidier to just pass the module, in my opinion.

(2) `latitude` and `longitude` are currently members of `SZACalcVars` module. I'm planning to combine those into `zenithData`, as that's fairly natural. I have that in a branch somewhere, I just haven't made a PR for that yet. I'll probably wait until everything else has been merged anyway, as that area of the code has various edits happening to it.

rs028:
Okay sounds good,
I am thinking, if it makes sense to have latitude and longitude in degrees (as it is now) and redefine lat and longt in radians. At the moment only lat is in radians and longt is equal to longitude, which does not make sense.

spco:
Perhaps the best way to fix this is to move the definition of `lat` after the `radians` definition, and on a single line each, e.g. `lat = latitude / radians` rather than its current use of multiple lines. Given that `longt` is not used in the formula, I'd actually just get rid of it.

On that point, surely `longitude` is needed for the `lha` calculation, given that the longitude will affect the local time? Or is the time data all given in the timezone associated to that longitude, not GMT?

spco:
Let me know when you want me to review this again - for the record, here I think we need `max( 0.0, cosx )` as max can't compare an int against a real - both need to be real.

rs028:
Okay, so I think if `lat` is just a subroutine variable, then we can get rid of `longt` and output `latitude` and `longitude` in degrees in `photoRateCalcParameters.output` which makes a lot more sense. Then merge `SZACalcVars` and `zenithData`, as you said. I can do it in this PR.

spco:
Ok, feel free to do that merge of the modules. Outputting in degrees makes more sense, yes.

rs028:
Should be all done, but wait to merge until I have checked the claculations of LHA and DEC.

Regarding longitude and LHA, it probably depends whether the model operates in GMT or Local Time, which I am not really sure to be honest. That's something to be checked. Maybe a separate issue?

spco:
Conflicts due to #197 now resolved

spco:
@rs028 what would you like to do here? I'm happy to fix these conflicts, merge this as it is, and have you submit another PR to check the LHA and DEC calculations, and check whether we're in local time or GMT? Would that work for you? It might be the simplest, and would give a clean slate for that checking.

rs028:
Okay, now how do I avoid the mess I fell into while trying to rebase the `press_m` branch?

spco:
@rs028 now this is merged, checking the LHA and DEC calculations and local/GMT time should be fairly straightforward :)

Issue #47, opened 2017-04-05
Extend tests to cover more .fac file examples


spco:

rs028:
I am thinking about rationalizing and extending the tests. At the moment they are a bit similar to each other and often test only specific parts of the code (sometimes very specific, for example `single_reac_of_interest`). 
I think each test may be used to test a few (unrelated, if possible) settings and/or initial conditions or outputs at the same time. Thoughts?

spco:
Yes, I totally agree. We should attempt to reduce the overlap between tests, although I wouldn't push that too far, in case we start to lose previous test coverage.

Looking at the tests we have in place, I think`short_extended` and `single_reac_of_interest` should be tweaked. The particular place we should be trying to cover is `environmentVariables.config`, which has lots of options, but the current tests all largely use the same choices.

As a first draft of changes to make, could I suggest:
(1) `short_extended` to use fixed `RH`, with `H2O = CALC`.
(2) `single_reac_of_interest` to use `DEC = CALC`, `ROOFOPEN = 0`
(3) leave most of the `spec...` tests as they are (don't want to change too many variables between tests) except perhaps change `spec_yes_env_yes` and `spec_no_env_yes1` to use a more helpful constrained variable - since `BLHEIGHT` is never used in any calculations, constraining it doesn't have much effect on the solution!

What do you think of those suggestions? Happy to talk it through more, that's just an initial suggestion.

P.S. as I've touched on in (3) above, none of the source code itself makes use of either `blh` or `dilute`. Taking a look at the .fac files, I don't see them using it either. Is it worth keeping these around in case of a hand-crafted .fac file that uses them? Or is there an option in the MCM to use these? I guess my point is that `full.fac` should contain just about all the reactions, and none use either of these. Happy to leave them in there if there is a possibility of them being used - otherwise they are just clutter!

rs028:
Sounds good. This is what I was thinking:

1) Test the single reaction of interest in `short` and get rid of that test
2) have a few `short_extended` tests to check all the `environmentVariables.config` options
3) the `spec_*` tests as they are now, with some tweaks
4) a `full` test with most options turned on

BLHEIGHT and DILUTE are meant to be used if you want to add processes to the base MCM model. I can add at least one of them in the `full` test.

What do you think?

spco:
1. `single_reac_of_interest` was created to check that behaviour seen in #34 and #81. I'd agree that this is fixed and unlikely to return, so removing the test is fine.
[...]

PR #203, opened 2017-06-02
JFac test

spco:

PR #204, opened 2017-06-02
Test scenarios

rs028:
This branch contains a reorganization of the test scenarios. As discussed in #47 the `short` and `single_reac_of_interest` tests have been merged, and given more sensible parameters.
The test `short_extended` has been converted into `short_ext1` , `short_ext2` , `short_ext3` , `short_ext4` to verify the calculations and conversions of environment variables H2O, RH, M, DEC, ROOFOPEN.

The other tests `spec_*` will need tweaking, but probably better in a separate PO (lots of files!)


rs028:
NB: short_ext2  and short_ext3 give different results for DEC. Need to investigate before merging.

spco:
This looks good - I'm just going to interject to say don't do anything to the `spec_*` tests, as you mentioned above, as I'm also working on tweaking one of those anyway (to get the JFac stuff working - the tests are currently being a pain because OSX and Linux are giving different values). Definitely better to do those after I've done mine, and definitely in a separate PR too - this one is very much pushing what's acceptable in size!

rs028:
Okay. I will finish this and then pass it on to you. I can't verify why it fails the tests on macos, though,

spco:
Ok. I need to do more work on understanding why Linux and macOS are behaving differently - the test in #203 is also giving different results, as does #128 which still has never been merged. Once I understand that, we can proceed with merging this.

spco:
@rs028 , can I ask whether the stoichiometric coefficient will ever be anything other than 1? This is currently hard-coded to 1 in `readReactions()` (`clcoeff` and `crcoeff` are the two variables that hold this). Is it ever likely to be anything other than 1? If not, my investigations are showing that I might do well to remove it to improve the macOS/linux difficulty. I'll need to investigate further, but Id like to know whether it _could_ be stripped out or whether that is not acceptable.

spco:
To add to the above, it seems that removing the stoichiometric coefficient doesn't make much of a difference to the numerical stability anyway, so my question is neither here nor there (although if it really won't be anything other than 1 then let's remove it!). I've discovered where the algorithm is going wrong, and it's a bit of a pain - it's numerical stability of subtraction of very similar numbers. I'll have to look into whether I can force it to be more robust. Otherwise, it might be enough to change the underlying .fac file in some way to sidestep the issue in the tests.

rs028:
I am not sure, to be honest. If it does not create a problem, then I would say to keep the stoichiometric coefficient, just in case.

Can you elaborate on the issue with the fac file?

spco:
Ok, we'll leave it for now, as it doesn't make a difference.

The issue (I'm still investigating to understand this further) seems to be an issue of numerical stability. Deep within the resid() function, we hit a line that essentially does the calculation a = -2.990519116362E+07 + 2.990519116361E+07. This is a well-known issue with subtraction in finite precision calculations - the difference between these two is minute relative to their size, and so the precision of the resulting number is affected, because the finite representation loses digits of accuracy.

It seems like this is the cause of the difference we see - linux and macOS are treating this very slightly differently, but then this difference gets multiplied over successive iterations (we're doing several thousand iterations so this quickly balloons).

The usual way to treat this is to rearrange the equation to avoid losing precision. I need to spend a bit of time seeing whether this is possible. Otherwise, it may be the case that this situation is in fact incredibly rare but we're unlucky to have hit it due to the numbers in our .fac file - perhaps changing one a little will be enough to avoid losing precision. The former option is preferable, since it would improve the code, but I need to see further how it plays out to see whether that's possible.

rs028:
I see. I will leave that to you then :)

spco:
I've just had to tweak the DEC value in `short_ext3` to give a stable value. This instability across platforms is a real pain, but I can't look at it right now.

rs028:
@spco All right. Good job. How serious do you think this instability issue is? I mean can we trust the results on a given platform?

spco:
We need to have a chat about this, as I think it is going to require further attention. I've not seen significant differences in the overall simulation outcomes in the tests we've run, it's usually just a case that it takes a different number of iterations, and outputs slightly perturbed values.

That said, it is a worry, especially as I don't have a firm grasp on when this might actually have a large effect on the overall outcome.

Changing to use openlbim has helped, but we still get varying results, and enough to worry me. I think it's probably linked to the conditioning of the system - we're handling some number of the size 1e15, and some of size 1e-30 - that is never a good situation to be in finite precision arithmetic, and so it will need more work to make this algorithm more robust. A very good starting point would be to work out where and when it issues IEEE underflow warnings, as that's a signal that precision is definitely being lost.

Can we talk in person about this sometime soon? Perhaps tomorrow?

rs028:
Yes, I am in all day tomorrow. Just send me an invite for a time that is good for you.

PR #205, opened 2017-06-04
Default config

rs028:
This PR contains only cosmetic changes to make it easier to read the code: tidying up spaces, introduce consistent separation beween subroutines and modules, wrapping comments, etc...
It also sets the default config files to the default values as per wiki pages. Ready to be merged.


rs028:
NOTE: should subroutines `FCVJTIMES` and `FCVFUN` (in `atchem.f90`) be moved to `solverFunctions.f90`?

atchem.f90
```diff
@@ -163,23 +162,30 @@ end subroutine FCVFUN
   ! Read in JFac species, and set jfacSpeciesLine
   call readJFacSpecies()
   write (*,*)
-
```
spco: Personally, I'd rather keep this blank line, as the lines below it are part of the next logical section.

atchem.f90
```diff
@@ -191,17 +197,22 @@ end subroutine FCVFUN
...
+  ! species, and the remaining elements being the numbers of the
+  ! reactions in which that species is a reactant
+  !
+  ! Fill the remaining elements of each row of reacIntSpecies with the
+  ! numbers of the reactions in which that species is a reactant
   call findReactionsWithProductOrReactant( reacIntSpecies, clhs, reacIntSpeciesLengths )
   write (*, '(A, I0)') ' reactants of interest (number of species found): ', size( reacIntName )
   write (*,*)
 
-
```
spco: Same again

atchem.f90
```diff
@@ -313,8 +337,7 @@ end subroutine FCVFUN
   write (*, '(A, I0)') ' setting maxstep ier = ', ier
   write (*,*)
 
-  !   SELECT SOLVER TYPE ACCORDING TO FILE INPUT
-  !   SPGMR SOLVER
+  ! SELECT SOLVER TYPE ACCORDING TO FILE INPUT SPGMR SOLVER
```
spco: This should remain on two lines, as SPGMR SOLVER is the first of the 3 choices.

configFunctions.f90
```diff
@@ -91,21 +101,24 @@ subroutine findReactionsWithProductOrReactant( r, chs, arrayLen )
     return
   end subroutine findReactionsWithProductOrReactant
 
+  ! ----------------------------------------------------------------- !
```
spco: Are these the width of your screen? You obviously have a different linewidth to me! And I suppose these are just visual aids for you to see the separation between functions? Does 2 blank lines not do this? Also, do you use a 'linter', which colours the keywords, which really helps with visualising the flow? I can't remember your setup.

parameter_modules.f90
```diff
@@ -25,35 +28,40 @@ subroutine set_solver_parameters( input_parameters )
...
-    ! This means FCVJTIMES() in solverFunctions.f90 should be used to approximate the Jacobian.
+    ! TODO: convert this to boolean?  If JvApprox==1 and
+    ! solverType={1,2}, call FCVSPILSSETJAC() below, with non-zero
+    ! flag.  This means FCVJTIMES() in solverFunctions.f90 should be
+    ! used to approximate the Jacobian.
     JvApprox = nint( input_parameters(3), SI )
-    ! This is never used, but is referenced in a comment in FCVJTIMES().
-    ! TODO: delete?
+    ! This is never used, but is referenced in a comment in
+    ! FCVJTIMES().  TODO: delete?
```
spco: Could you keep the TODO here and on line 48 on separate lines?

parameter_modules.f90
```diff
@@ -110,29 +121,31 @@ subroutine set_model_parameters( input_parameters )
     write (*, '(A)') ' Reading model parameters from file...'
     interpolationMethodName(1) = 'piecewise constant'
     interpolationMethodName(2) = 'piecewise linear'
-    ! maxNumTimesteps sets the maximum number of timesteps to calculate.
-    ! Calculation will terminate when currentNumTimestep>=maxNumTimesteps.
+    ! maxNumTimesteps sets the maximum number of timesteps to
+    ! calculate.  Calculation will terminate when
```
spco: Extra space not needed.

parameter_modules.f90
```diff
@@ -110,29 +121,31 @@ subroutine set_model_parameters( input_parameters )
...
-    ! Use the local variable speciesInterpolationMethod to set the value speciesInterpMethod,
-    ! the private member of MODULE interpolationMethod.
-    ! getSpeciesInterpMethod() is called by getConstrainedQuantAtT.
-    ! Values:
-    ! 1: Piecewise constant
-    ! 2: Piecewise linear
-    ! otherwise: error
+    ! Use the local variable speciesInterpolationMethod to set the
+    ! value speciesInterpMethod, the private member of MODULE
+    ! interpolationMethod.  getSpeciesInterpMethod() is called by
```
spco: getSpeciesInterpMethod() should begin a new line.

spco: Also the different values below should have separate lines

solverFunctions.f90
```diff
@@ -131,8 +140,8 @@ subroutine mechanism_rates( t, y, p )
     use utilityFunctions_mod, only : calcAtmosphere
     implicit none
 
-    ! calculates rate constants from arrhenius information
-    ! output p(:) contains the rate of each reaction
+    ! calculates rate constants from arrhenius information output p(:)
```
spco: information. Output p(:) (new sentence)

spco:
Is https://github.com/AtChem/AtChem/wiki/Data-files the wiki page you mean when you talk about the default config files? That page needs updating as it doesn't reflect reality very much! e.g. the concentration file doesn't have the -1 line and onwards, and there's mention of zip files containing data files.

Finally, subroutines `FCVJTIMES()` and `FCVFUN()` do need to stay in atchem.f90. Most libraries work by just knowing what the interface of each function looks like, and not worrying too much about the internal contents of the functions. In this case, _we_ provide the internals of the FCVFUN and FCVJTIMES functions - all we have to do is ensure that their interface (that is, their inputs and outputs) look like the CVODE library expects. Then, when the CVODE library looks for this file, it just has to find something that matches those characteristics.

In order for the interface of our user-provided FCVFUN etc to be matched against what CVODE expects, CVODE has to at least be able to see the interface. Placing this inside a module confuses things, so the library can't find the interface. Instead, we just define the functions as usual, but outside of a module.

On top of this, just to confuse things, the fact that these functions use assumed-size array (we don't want to have to pass the vector sizes along with the vectors) requires the functions to have an _explicit interface_. Hence, we add the lines 60-101, which tell the compiler what the interface of the functions looks like, but without defining the internals. We then leave the internals until the bottom of atchem.f90.

So to summarise, because CVODE looks directly for these functions, they can't be in a module. And because they use assumed-size arrays as arguments, they need an explicit interface.

Hope that makes sense!

configFunctions.f90
rs028: I use emacs, which has a fortran mode. I think emacs sets default line width to 72 for historical reasons. Personally I like a visual aid to identify the functions and subroutines. But all these things are a bit personal preferences and i am happy to change them to be more standard!

rs028:
Yes it makes sense.

The default config refers to the environment variables page actually.  I though it would make more sense to have those values in the config files). I will apply the changes you suggest when i get to the office.

configFunctions.f90
spco: Ok. Ultimately, it's your code so what works for you is best! Do you always have a 72-character window in emacs, or do you use something wider? If so, I'd make it as wide as you use - modern displays are not limited to 72 chars before they become unreadable. If the visual aids work for you, then I won't push to remove them.

rs028: I tend to use 72, although its is not a hard limit. It is true that the comments require a bit of flexibility for clarity. While we are on the aubject ahould every module and subroutine have a header?  (Not to add to your workload, i am happy to do these things as it helps me understand the code better)

spco: Up to you - if you want to do that, and it will help you or other users to understand what the module or its contents do, then I'm all for it :)

spco:
I'm happy to merge this if you are - any module headers can come in separate PRs - are you happy for me to merge?

rs028:
Yes. thanks.

spco:
Thanks for this @rs028

PR #206, opened 2017-06-07
Tidy comments

rs028:
I am adding and tidying up comments and headers to the f90 files. This is work in progress, but I am already opening a PR so you can review the changes a little at the time and let me know if I do something which is not good practice. Is that okay?


rs028:
The first commit is changing the name of parameter_modules.f90 and ro2_rates.f90 to make it consistent with the other filenames (rule being camelCase for regular f90 files, lower case and dash for generated f90 files).

spco:
This generally looks alright, although ro2-rates.f90 is a generated file so needs removing from this PR - changing the name within the python generating script should be enough.

rs028:
Yes of course.

I am trying to set a style with a general comment of what a subroutine does on top (`! ???` when missing or not clear) and more specific comments through the code. But I don't want to go around in circles, so tell me if I am undoing things you have already done.

spco:
Thanks for this. I'll have to go back and fill in all the ??? comments in another PR - it's useful to have them labelled in this way, so I find more of them!

rs028:
That was the idea ☺

Issue #16, opened 2017-03-21
JFAC crashes when calculated from constrained jNO2


rs028:
JFAC is a scaling factor for the photolysis rates.  It is the ratio between the measured and calculated photolysis rate of NO2 (jNO2 or J4) to adjust for cloudiness.
If the jNO2 is constrained, then JFAC can be calculated from the constrained value vs. the calculated J4. Alternatively it can be one of the constrained environmental variables (in environmentalConstraints/). Or it can be set to NOTUSED and ignored.

It works fine if constrained, but it crashes the code if it needs to calculated from constrained jNO2. It is not clear if it is enough to set JFAC as CALC in modelConfiguration/environmentVariables.config or if J4 needs to be specified in JFacSpecies.config as well.

spco:
Note: there's now a new wiki page called 'Photolysis rates and JFac` detailing the palgorithm. I'm still working on a fix and test for this.

spco:
I think this can be closed - we require JFAC to be set to CALC, as well as J4 being in JFacSpecies.config, and there being a file entitled `J4` in the `speciesConstraints/` directory.

rs028:
This still does not work very well.

spco:
Is JFAC always calculated from J4/jNO2? I believe that the current implementation allows any photolysis rate to act as the JFAC species, so long as it has the parameters defined in `photolysisRates.config` and the constrained data in `constrainedPhotoRates.config`. Should it only be allowed for J4?

And is there any chance you could give a demonstration of this failing? This may have been fixed by the photolysis/JFAC tidyup recently, of course.

A comment above has highlighted the wiki page related to this - I'll update that page to give a robust description of how photolysis and JFAC work together.

rs028:
In principle yes, any photolysis rates could be used to calculate JFAC. In practice, J4 is the most likely to be used, because it is one most commonly measured. so I think it is fine to leave the possibility to use others, and set J4 as default choice.

I will have to dig up some to find an example of it failing. In general I think the problems tend to occur when there are zeros or a NaNs in measured jNO2. In part this is related to the discussion in #276, i.e. whether the photolysis rates should be set to zero at night (and hence JFAC to 1).

PR #207, opened 2017-06-08
Copy jfac

spco:
This contains all the bits of the jfac branch that I think probably shouldn't break the tests.


spco:
Related to #16 and #203 - I will now rebase #203 on top of this, and try to figure out what on earth is going on!

PR #208, opened 2017-06-08
Jfac split

spco:
~~Splits some of the JFac_test branch out, to just test the spec_yes_env_no test, without adding any new tests.~~


spco:
This PR adds openlibm support, which is now necessary to run the tests. In order to fix the testing regime, I've had to loosen the numdiff tolerances a little, and tweak some of the tests.

@rs028 - once you update to this version, you will first of all have to run 
```
./tools/install/install_openlibm.sh .
```
when you are in the main AtChem directory - note that the dot on the end of that line is vital! This will install openlibm in the directory `openlibm-0.4.1/`. This is a library containing implementations of various mathematical functions, e.g. exp, pow and cos. Linux and macOS use different versions of this by default, so by installing this here, we can force AtChem to use this one version, and reduce the differences seen between Linux and macOS.

rs028:
@spco All right. I am doing a bit of housekeeping (comments, cosmetic changes etc...). It is not super important and it can be split in several PR. Since it touches all the f90 files, when is it better to rebase/merge?

spco:
I would say hopefully I will be done with this (I still need to add the two tests with the JFAC) by the end of the day - would you be able to rebase your work on top of that? I can let you know when I'm finished.

spco:
If you are just doing cosmetic changes, and it touches more than a couple of places in each file (e.g. if you're adding comments to multiple functions in a file) then splitting it up into smaller PRs would be helpful, and more manageable for you.

rs028:
Should be fine.

spco:
Ok, I hadn't realised how much I'd need to tweak - I need to merge #128, #203 and #204, which I won't get done today. If you want to make progress today then you can start work from the current master (as I shouldn't be touching many of the .f90 files anyway), otherwise I'll be working on this first thing tomorrow.

PR #209, opened 2017-06-09
Fix dec

rs028:
This should fix the calculation of the Sun Declination. I will add comments if everything check out.
Next is to fix the Solar Zenith Angle. I can do in this PR or in another one if you are keen to merge this now.

I have moved calcAtmosphere (again) I think it makes more sense to have calcM and calcAtmosphere together and the subroutines dealing with solar angles in their own module. If we go that way, the modules should probably be renamed as well.


spco:
This won't build because `calcDec` is expecting `t` as input as it's defined with `intent(in)`, but isn't an argument. Shouldn't `calcDec` take the time of day into account, as it's a continuous function based on the progress through the year? As it stands currently it looks like it has a set value per day, which probably isn't desired?

rs028:
I forgot to take that out. The declination depends on time, but only on the day and month, not hour.
The `dayOfYear` variable in the new calcTheta function should account for that. I am also going to add theta to the output file `photoRateCalcParameters.output`

spco:
Surely declination angle changes throughout the day? It's a continuous function, and is different at midnight to midday on the same day. The equation used to come from https://en.wikipedia.org/wiki/Position_of_the_Sun#Declination_of_the_Sun_as_seen_from_Earth, the 3rd equation in the Calculations section.

spco:
My point is that the new formula must be implementing a step-function to approximate, but wouldn't it be better to keep the more accurate formula?

rs028:
Yes, it does, a little (I think ~0.07 deg over 24 hours). So this is like saying that we have one value of DEC for each day, The purpose of DEC is only to allow the calculation of the SZA, so this could be an acceptable approximation.
I am not opposed to a continuous function, but that formula does not give correct results, when checked against the NOAA calculator at https://www.esrl.noaa.gov/gmd/grad/solcalc/
I suspect astronomers define this a bit differently but I am not 100% sure.

spco:
Ok. Does the new formula give better matching to the NOAA calculator? Looks like the formulae used are in the spreadsheets available at https://www.esrl.noaa.gov/gmd/grad/solcalc/calcdetails.html, which might be useful to look at.

rs028:
Yes, that's what I am going to do next.

rs028:
I am struggling a little here, as theta is not being calculated as it is supposed to be and I suspect this is because of subroutine `calcDateParameters`. It is called but it gives these results (with the default date 21/6/2010:

    dayOfYear                    -85
    dayAsFractionOfYear   0.0000000000000000

Which seems odd. Maybe I inadvertently changed something. Do you mind checking real quick that you get the same numbers?

spco:
I haven't tested this myself, but when are you printing these numbers? `calcDateParameters()` is called on about line 233, has that not been called when you print these?

rs028:
I am printing from calcTheta, so it should use the values stored in the `date` module, if I understand how it works.

spco:
Yes, it should. Could you try adding the same output lines at the end of `calcDec()`, and check whether that (a) is called before `calcTheta()` and (b) gives the same or the correct answer?

rs028:
Something is not right. I added a write instruction at the end of calcDateParameters itself. and got that output. I think I may have found a little bug.

spco:
Ok. Either make a new issue, or feel free to investigate this. I'm currently down the rabbit-hole of this JFAC stuff - I now understand what's going on, but am still working on a solution. It's not pretty!

rs028:
The calculation of DEC is now okay (see attached plot). Working on LHA and SZA, now. 

dec_calc

spco:
Looks good. I'll review this as it is, and then LHA and SZA can either come in another PR or be appended to this, whichever you would prefer.

utilityFunctions.f90
```diff
@@ -1,42 +1,51 @@
...
-  end subroutine calcAtmosphere
+  end subroutine calcTheta
 
-  ! ----------------------------------------------------------------- !
-  ! calculate the Sun Declination (radians): the sun declination is
-  ! the angle between the Sun and Earth's equatorial plane
-  pure function calcDec( t ) result ( dec )
+  ! -----------------------------------------------------------------
+  ! calculate the sun declination (radians)
+  pure function calcDec() result (dec )
```
spco: space in `( dec`

constraintFunctions.f90
```diff
@@ -149,6 +149,9 @@ subroutine getEnvVarsAtT( t )
     character(len=maxEnvVarNameLength) :: this_env_var_name, orderedEnvVarNames(size( envVarNames ))
     logical :: pressure_set, rh_set, temp_set
 
+    ! TODO, is this the best place to call calcTheta?
```
spco: This looks like a great place - the only functions that use `theta` are `calcDec()` and `calcZenith()`, and they are only called below.

spco: Actually, I'd just move it to below the large comment below - it risks getting lost up here, as it's the only non-declaration statement before the comment.

atchem.f90
```diff
@@ -218,8 +218,8 @@ end subroutine FCVFUN
   call set_model_parameters( getParametersFromFile( trim( param_dir ) //  "/model.parameters" ) )
   write (*,*)
 
-  ! Set the members dayOfYear, dayAsFractionOfYear, secondsInYear of
```
rs028: I would like to fix everything in one go.  I just realized that theta is not calculated the right way (or rather it is, but if a value for theta is not used for the whole day the equation of time gets screwed up and so does the SZA. I think if theta is corrected then everything else will fall into place.

dataStructures.f90
```diff
@@ -64,18 +64,26 @@ module date
...
+    integer(kind=SI) :: monthList(12), i
 
     monthList = [31_SI, 28_SI, 31_SI, 30_SI, 31_SI, 30_SI, 31_SI, 31_SI, 30_SI, 31_SI, 30_SI, 31_SI]
-    ! calculate which day of the year day/month refers to
-    dayOfYear = sum( monthList(1:month - 1_SI) )
+
+    ! calculate day of year from day/month (set in model.parameters)
+    ! January 1 = 0, January 2 = 1, etc...
+    dayOfYear = 0
+    do i=1, (month - 1_SI)
```
rs028: I don't know :(
I could not get it work using `sum`, for some reason, that's why I changed to a loop.

atchem.f90
spco: Ok. So long as we end up with a value of `dec` and `zenith` that changes continuously, I'm happy.

dataStructures.f90
spco: That's rather odd. Can you try changing it back, and, if it doesn't work, submit what you have for me to inspect? Returning it to its original shouldn't break it!

constraintFunctions.f90
rs028: Btw, that large block of comment contains a lot of info. part of it is explanation of what the subroutine does and part are details. so I think it should be split up and some info moved to the header (where ??? is). Do you agree?

spco: Potentially, yes - I'll fix that.

rs028: Question. Sometimes modules are just called. Other times there is a `only` instructions that specifies which parts of the module are needed. Is there any advantage to do that?

spco: If only a single member (or only a few) is used, then `only` is useful to ensure other bits aren't accidentally used (part of what's known as defensive programming). That's the usual criteria - if most or all members are used then just use the whole module without `only`. However, in some of the functions I've also used `only` as a useful way of documenting where the different variables come from - for example in `readPhotolysisRates()` and `readPhotolysisConstants()` I've added a long list to the line `use photolysisRates_mod, only : ...` because that's helpful to show where most of the variables are coming from - if it's not explicitly stated then it can be a bit opaque. It also has the added benefit of defensive programming as above, but that's less of the focus.

If it's fairly obvious, and not too much risk of inadvertently using member variables incorrectly, there's no need to add `only`.

rs028: Makes sense. Should the code be reviewed to make sure it is consistently applied, or is it something you already did?

spco: I've not done a full review of it, but it's generally in a good enough state - I don't think there's any need to improve it.

rs028:
Okay. I need to double check against the NOAA calculator, but it should now be correct.
I left the papers at home, so I will add the comments and documentation later. If you are keen to merge to see if it impacts the numerical issues you are having, go ahead. I will open a new PR later to tidy up and document.

spco:
Ok. I'll leave this until it's completed. I've worked out the issues with the numerical issues, and am working on a fix, so this shouldn't impact that. Expect a fix before the day is out! 😃

rs028:
Both DEC and SZA are now correct. Checked agains the NOAA calculator at https://www.esrl.noaa.gov/gmd/grad/solcalc/. Comments and references need to be added.

zenith_angle

**Note**: The equations assume that time is in GMT. Does this affect other parts of the code? I think not, but need to check. This info needs to be added to the wiki.

**Note2**: a test scenario should be added, with different lat/long.

spco:
Is this ready for review?

rs028:
Yes.

spco:
Looks all fine to me, apart from one question - what kind of period do you do this modelling for? Less than a day, a few weeks, or over a year?

rs028:
Usually no longer than 2-3 months. Why?

spco:
I ask because the variable `day`, `month`, `year` all hold the _initial_ day, month, year. Using `year` to then decide whether we're in a leap year or not is only valid so long as the model doesn't go outside of that year. So it's OK to start a model on 1st February and run it for a couple of months, but if started on, say, 31st December 2012, then `year` will always hold 2012, and so the model will always think it's a leap year, but by the time it hits the end of February, 2 months later, the real year is 2013, which isn't a leap year.

rs028:
Ah yes, that is not the only problem if the year changes during the model run. The calculation of dayOfYear - and everything that depends on that - is also going to be affected, because Jan 1 should be day 0 not day 366 (or 367 if leap year).
Probably not urgent to address now, unless you cna think of a super quick way to do it, but I can open a new issue about that.

spco:
Leave it as it is, and open an issue - better to get this merged. Do you mind if I rewrite the history of this PR? I probably don't want to squash it into a single commit, but the history isn't very helpful in its current form.

rs028:
Go ahead! :)

spco:
While I'm here, can you give me the error message that comes if you use `sum` for the days rather than the `do` loop? And also the compiler version you're using. I'd like to understand why that isn't working, although I'll save that for a separate PR.

rs028:
Ah yes, forgot about that. It's not giving an error. It just gives the wrong result, and I can't figure out why.

I am using gfortran 4.8.5

spco:
Ah, I see. I'll leave it for now, could you just open a new PR for that? If you can give me an example of where the two give different results that would be great.

spco:
Merged! Thanks for this work Roberto. I'm working hard on fixing the numerical issues (still!) but will get there.

rs028:
Thank you! You are doing all the hard work! I though i might do a bit more of tidying up and commenting, unless it is going to make more difficult merging what you are doing or there is something else more inportant I may help with.

PR #210, opened 2017-06-09
Linux jfac split

spco:


spco:
This is not to be merged, #208 supercedes this.

Issue #201, opened 2017-05-31
Correct handling of negative and zero angles


spco:
When there's no light, the trig functions in the formula for the photolysis get cut off at zero. To avoid a division by 0, we currently set these to 1e-30, but it might be better to properly handle this situation instead. Worth investigating.

rs028:
Note: the 1e-30 hack appears in other parts of the code (eg, subroutine `calcZenith`).

spco:
Fixed by #218

PR #218, opened 2017-06-16
Handle small cosx better

spco:
Add a 'infty_secx' option which is triggered when cosx is less than a given tolerance, so that we treat small cosx correctly/robustly.


spco:
Fixes #201

PR #225, opened 2017-06-16
Move JFAC species declaration to environmentVariables.config

spco:
Change to set JFAC species name within environmentVariables.config. Still need to test this, and remove all the JFacSpecies.config files.


spco:
This PR changes the input mechanism so that, in the calculated-JFac case, the JFac species is input as a name in `environmentVariables.config` rather than in the input file `JFacSpecies.config`.

Added two tests - `spec_yes_env_no_with_jfac_fail1` is designed to fail, as it is missing the required input file `J4` in `environmentConstraints/`. `spec_yes_env_no_with_jfac_fixed` uses a fixed jfac value.

Issue #211, opened 2017-06-13
Change of year


rs028:
If `year` changes during the model run, the variable `dayOfYear` should reset and the code should check if the new year is a leap year. This is to avoid `dayOfYear` to run over 365/366 and the model to think every year is a leap year (or viceversa).

PR #226, opened 2017-06-20
Handle leap years

spco:
Handle leap years in a sensible way, by separating out the start date and time from the current date and time. Add a new test, short_end_of_day, to test this works as expected.


spco:
Fixed #211

PR #235, opened 2017-06-21
Edits to .travis.yml

spco:
Remove brew update. Convert precise to sudo: false, and add a sudo: required trusty build.


spco:
Deleting this branch, as `sudo: false` makes the `-Wpedantic` argument fail - not sure whether this is the real issue, or whether the lack of sudo means gfortran isn't installing fully.

PR #240, opened 2017-06-22
Make build.sh more flexible

rs028:
Just an idea, to make it possibile to run build from the `tools` directory rather than the main directory.
Also, the 3rd argument is not specified.

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -1,24 +1,29 @@
 #!/bin/sh
 #
-# $1 is the input file
-# $2 is the directory in which to place the mechanism-rate-coefficients.f90 and
-#    mechanism-rate-declarations.f90 files. This, by default, points
-#    to the base directory of the source files of AtChem.
-# $3 is ...
+# Script to ....
+#
+# $1 is the mechanism input file (*.fac)
+#
+# $2 is the directory in which to place the output files:
+#    - mechanism-rate-coefficients.f90
+#    - mechanism-rate-declarations.f90
+#    By default, it is the base directory of AtChem.
+#
+# $3 is the directory in which to place the output files:
+#    - mechanism.prod
+#    - mechanism.reac
+#    - mechanism.species
+#    By default it is modelConfiguration/ in the base directory of AtChem.
 
 set -e
 echo ''
-echo $1
-echo $2
-echo $3
+echo "input file:" $1
+echo "output directory:" $2
+echo "output directory:" $3
 
-if [ "${PWD##*/}" == "tools" ]; then
-    echo "call ./setup_atchem.py"
-    python ./setup_atchem.py $1 $2 $3
-else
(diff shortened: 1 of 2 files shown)


spco:
Rather than using if-else, could you explore using `python ${BASH_SOURCE%/*}/setup_atchem.py $1 $2 $3` ?

And you could add that $3 is the directory in which to place the output files `mechanism.prod/reac/species`.

rs028:
Will do. 

I am also adding back the `multiplot.py` script. It was in the `web_tools` which we removed in the beginning, but I think it may be a good idea to keep in the tools to allow a quick&dirty look at the model results. I am not sure if it works, actually, but I will start adding the file to this PR.

spco:
Ok. If it doesn't work immediately, I'd recommend putting it into a separate PR as it's independent of `build.sh`, but if it does work then it's probably ok to put them both in one PR.

rs028:
Using `BASH_SOURCE` works but then it does not know how to be in the main directory to do `make`.
Also I am not sure about the directories being defaults in the explanation of the arguments.
Sorry I was tryin gto simplify things and may have opened a can of worms here.

spco:
Ah yes - make does have to be called from the correct directory. In general, trying to write build scripts like this to handle multiple situations is a rabbit-warren, so it is probably better to just document that all these scripts should be run from the main directory and leave it there?

spco:
I think the defaults are ok, so long as the script is called from the right place - `setup_atchem.py` fills the arguments with default values if they're not provided.

rs028:
Okay. I will just revert to the original and just leave the leave in the comments then.

One question: I see there is a commented NOY section in `mech_converter.py`. Is this a leftover from the previous version ar has it been rewritten together with the rest of the script?

rs028:
BTW, I think I found out why git thinks that `tools/mcm_example.fac` has been modified after build.sh is run. It has to do with newline characters in the middle of a chemical equation in the original mechanism file. For example lines 262-265 in https://github.com/AtChem/AtChem/blob/master/tools/mcm_example.fac

spco:
On the `NOY` front: the current PR #238 about python removes that, as it was just a copy of the RO2 part. It can always be re-added in the same manner if it's ever needed.

Yes, `build.sh` calls `mech_converter.py`, which calls `fix_mechanism_fac.py`, which cleans up the .fac file to fix those troublesome newlines. I think that I left the original file with those problems in it, so that we would see that the `fix_mechanism_fac.py` works. Perhaps we should replace it with a pre-processed version anyway, just to avoid the irritation of git thinking it's changed. I'll open a small PR.

rs028:
Stupid typo in the name of the python script, but I am on a train now. Are you pkay to fix that and merge? Othetwose I will do it tomorrow.

spco:
Sure, I'll do that.

Issue #246, opened 2017-06-23
Copyright/credit


spco:
Ideally all files should have a copyright/credit section in the header, which gives a few details of who wrote it and when.

This is made trickier by the fact that large amounts of the code was written beforehand, and we may not be able to trace who did what. Perhaps we then go down the route of adding all names to all the files, unless they are explicitly new (such as `parameterModules.f90`, and perhaps `solarFunctions.f90`/ `atmosphereFunctions.f90` given how almost all the functions have been re-written). In particular, the build and test systems are all our own work, so that should have a header stating such. 

This all helps to identify who did what later!

spco:
The usual convention is to add authors as they touch individual files, and to update the copyright year each time a file is edited.

rs028:
@spco, apart from copyright, credits, licence is there something else or are you done with all we talked about?

spco:
I think that's everything on my list! Once the license is in place, from my POV we are good for a release version.

spco:
Closed by #249

Issue #243, opened 2017-06-22
License


spco:
We still don't have a license in place. We definitely should.

spco:
@rs028 : In response to your question about MIT vs GPL, it seems that GPL requires any derived software to also use a GPL license, whereas MIT puts no condition on that, merely that the attribution is kept. I think MIT better suits our purposes for this, as we're more concerned with attribution than restricting further use - does that sound ok?

rs028:
Okay. I really don't have strong opinions on that, as long as it is a full open source. If you prefer MIT I'm happy with that. We could put to options to vote to the others indicating our preference.

spco:
Ok, I'll prepare two versions, and a brief description of the two choices. I'll email you before the day is out.

spco:
Closed by #249

PR #249, opened 2017-06-28
Add LICENSE.md, and a license header to each source file.

spco:


spco:
Addresses #243 and #246. These can both be closed when this merges. 

@rs028 Let me know if this looks OK to you - I have added copyright for you just you and I to all of the testsuite stuff, as that is all brand new.

spco:
This is ready for merge if you are happy with it @rs028 . The only remaining step is to rename it to AtChem2 or similar, and release the version.

rs028:
@spco, okay let's do the final merge! for the name I'd keep it simple. How about `Atchem+`?

spco:
My only issue with that is that GitHub doesn't allow `+` as a character in the repository name, so it would come out as `Atchem-`

rs028:
AtChemX?

rs028:
or back to the original idea of AtChem2?

spco:
Either of those sound good - your choice!

rs028:
Okay, let's go with AtChem2. Not very imaginative, maybe, but easy and clear.

spco:
Ok, sounds good. I think the simplicity is the most important part. I will do that tomorrow. If we change the repo name to match that (probably best long-term) there may be a little work to tweak your local repo as well, but we can sort that out.

rs028:
All right. I don't mind if I have to delete my repo and fork again. If it makes things simpler.

PR #250, opened 2017-07-07
bits and bobs

rs028:
This is just some comments I had forgotten and some formatting I came across while working on the wiki.


rs028:
@spco Could you have a look at https://github.com/AtChem/AtChem/wiki/2.3-Solver-Parameters
I was tidfying it up and there are a couple of parameters in there which actually don't appear in the `solver.parameters` file. I move them at the bottom of the page. Should they be in the file?

spco:
Looks good - I've just reverted the one .gitignore line, and will merge when this gets through.

spco:
And on the solver parameter issue, these were removed in #221, so can be removed from the wiki - good catch 👍

rs028:
Re .gitignore, doesn't the same logic applies to cvode and numdiff? should those be added to .gitignore as well?

spco:
That's a fair point. Please add those too. Changing that using the online editor is probably easiest.

PR #251, opened 2017-07-07
Change name to AtChem2

spco:
Change every occurrence of 'atchem' to 'atchem2', including changing file names.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,15 +5,15 @@
 #
 # Copyright (c) 2017 Sam Cox, Roberto Sommariva
 #
-# This file is part of the AtChem software package.
+# This file is part of the AtChem2 software package.
 #
 # This file is covered by the MIT license which can be found in the file
-# LICENSE.md at the top level of the AtChem distribution.
+# LICENSE.md at the top level of the AtChem2 distribution.
 #
 # -----------------------------------------------------------------------------
 
 #!/bin/sh
-# Makefile for AtChem project
+# Makefile for AtChem2 project
 
 .SUFFIXES:
 .SUFFIXES: .f90 .o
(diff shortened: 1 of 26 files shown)


spco:
This will fail tests, as the name of the repo hasn't been changed yet.

spco:
@rs028. I think I have now completed the transfer to the name AtChem2, including changing it in the wiki, and renaming the repository and organisation.

For you to do:
- Rename your GitHub fork to AtChem2 - go to Settings on your fork.
- Change your local repo: update to the latest version, rename the folder to atchem2, and update your remotes:
```
git remote -v
```
then for each remote, e.g. `origin`, do 
```
git remote set-url origin https://github.com/rs028/AtChem2.git
```
or for the upstream,
```
git remote set-url upstream https://github.com/AtChem2/AtChem2.git
```
After that, update any bookmarks or anything you have, and we're done! The old links should continue to work, but it's best to update them as another user could potentially make a new one with our old name! Any questions, shout.

Makefile
```diff
@@ -38,7 +38,7 @@ FORT_COMP    = gfortran
 CVODELIB     = /path/to/cvode/lib
 else
 FORT_COMP    = gfortran
-CVODELIB     = /path/to/cvode/lib
+CVODELIB     = /Users/sam/ReSET/Sommariva/cvode/lib
```
rs028: why has this been changed again?

spco: My bad...

spco: Fixed.

rs028:
All looking good. Just one question: Would it make sense to keep the name of the organization as AtChem and only change the name of the repo to AtChem2? I am thinking that maybe in the future the web interface, if they will want to rebuild it, could be hosted in the same organization in its own repo. What do you think?

spco:
I'm loathe to do that right now, as I've just changed it, but it sounds a good idea if they do want to do that in the future. For now, the GitHub organisation doesn't include AtChem online, so it might be less confusing if the name doesn't suggest otherwise?

rs028:
Fair enough. For my reference, if I need to do that in the future, changing the name of the organization is done via the web interface (settings) and then updating the git remotes? or is it more complicated?

spco:
Yep, it's as simple as that. 😄 

https://help.github.com/articles/renaming-an-organization/

PR #255, opened 2017-10-27
Header of time in envVar.output

rs028:
Use same naming convention for `time` in all output files.


spco:
@rs028 : I've fixed the build system (SUNDIALS have released a new version and the existing download is no longer available, so I've tweaked the script - see #256 ). I've restarted your CI tests above, so hopefully they will pass this time :)

rs028:
@spco  Thanks! I was going to do it myself, but had not found the time yet!

PR #262, opened 2018-01-16
Edit tools/install/install_cvode.sh

spco:
to correctly handle a relative install path. Also improve a couple of error messages.


spco:
Requires #264 to be merged, then rerun tests.

PR #266, opened 2018-03-27
Correct dec

### LINKED ISSUE #14: calculation of DEC is likely wrong. Check that SZA is correct.

rs028:
This should fix #14. Can be merged but there is still a minor bug in subroutine `calcZenith` with the calculation of variable `currentFracHour` (doesn't affect the results significantly as fas as I can tell).


src/solarFunctions.f90
```diff
@@ -109,7 +110,12 @@ subroutine calcZenith( t, dec )
     ! The local hour angle is the angle between the observer's
     ! meridian and the Sun's meridian. Time must be in GMT/UTC and
     ! longitude in degrees.
-    currentFracDay = currentDayOfYear + ( t / 86400.0_DP )
+    !
+    ! TODO: currentFracDay is wrong because t/86400 is not the
```
spco: So what's wrong here? The fact there aren't exactly 86400 seconds in a day?

rs028: No I think that `t` is not the number of seconds. What happens is that `currentFracDay` increases too rapidly. For example:

currentDayOfYear    currentFracDay
1                                 1.8
1                                1.9
2                                2.1
2                                3.2

You notice that only if the model is run for more than one day.  It's really just fractional part we are interested in, so if there is an easier way to calculate `currentFracHour` then we can get rid of it (Hope it makes sense)

rs028:
@spco I guess this cannot be merged until PR #264 is done, right?

spco:
Yeah, I will try to get to that at some point soon as it will hold us up.

rs028:
Okay. I can keep working with the current version, not a problem.

spco:
@rs028 : good news - I have fixed the numdiff issue from #264 . I've also pushed through the cvode change in #262 at the same time. At your convenience, rebase on top of the latest master version and repush to get this PR tested. 😄

rs028:
@spco something went wrong when I tried to rebase :(
I did
    git checkout master
    git pull upstream master
    git checkout correct_dec 
    git rebase master
    git push --force

No error messages, but it seems the merged version has not appeared on the branch on github (?)

spco:
I expect you needed a `git fetch upstream master` before `git pull upstream master`, so master will not have updated. Can you show the output of `git log --oneline --branches --graph` before doing that? That should show whether the master branch has updated locally.

rs028:
It looks like it did not. the last commit in the log is edf9534.
So to be clear the procedure is:

    git checkout master
    git fetch upstream master
    git pull upstream master
    git checkout correct_dec 
    git rebase master
    git push --force

is that right?

spco:
Yes, that should work. I'd take a look at the git log along the way to check things make sense e.g. before and after the rebase.

spco:
And check that the whole tree looks ok on your local copy before doing a force-push, of course!

rs028:
No. Something is wrong.  My `master` is even with `AtChem:master`, but the branch `correct_dec` is two commits ahead. However when I try to do `git push` (with or without the `force` option) I get the message that everything is up-to-date (?). Do IO have to issue a new PR?

spco:
No, that shouldn't be necessary. Can you send my the output of `git log --oneline --branches --graph` and `git remote -v` please?

spco:
Hold on, this is actually alright - everything is as it should be. Try hard-refreshing this page to see that the commits have been moved to after I commented yesterday. Looking at the Travis output, it ran 3 hours ago with the updated `master`, it just failed one of the tests on the Mac build.

rs028:
Okay, so, should I leave this open until there is time to sort it out? It is not a problem to do so, as there is still the minor issue with currentFracHour to sort out.

spco:
When this comes to merging, we'll need to redo the test outputs, as #280 fixed a bug that affects the output.

rs028:
@spco it took me the best part of the afternoon, but I think it is done. However:
1. it fails the macos tests
1. it creates `mechanism-rate-coefficients.f90`, `mechanism-rate-declarations.f90` and `ro2-rates.f90` in `src/` as well as in `src/gen` and I cannot figure out why.

spco:
On 2, you'll have to make sure that your Makefile is up to date:
```
diff Makefile tools/Makefile
```
should only show differences related to paths to CVODE and OPENLIBM and possibly FRUIT.

On 1, I'll see if there's anything we can do.

src/outputFunctions.f90
```diff
@@ -103,18 +103,18 @@ end subroutine outputSolverParameters
...
     logical :: first_time = .true.
 
     if ( first_time .eqv. .true. ) then
-      write (59, '(100A15) ') 't', 'LATITUDE', 'LONGITUDE', 'SECX', 'COSX', 'LHA', 'SINLD', 'COSLD', 'THETA', 'EQTIME'
+      write (59, '(100A15) ') 't', 'latitude', 'longitude', 'secx', 'cosx', 'lha', 'sinld', 'cosld', 'eqtime'
       first_time = .false.
     end if
 
-    write (59, '(100 (ES15.6E3)) ') t, latitude, longitude, secx, cosx, lha, sinld, cosld, theta, eqtime
+    write (59, '(100(1P e15.7)) ') t, latitude, longitude, secx, cosx, lha, sinld, cosld, eqtime
```
spco: Could you change the print format back to ES15.6E3 please? This is a regression from the rebase.

rs028: I was afraid it may happen. Can you spot any other instance where I undid something or should I go back and re-rebase the whole lot?

spco: No, that looks like the only one. Unfortunately in the meantime I have merged another PR and made yours conflict again, sorry.

rs028:
Should be okay now.

I checked `Makefile` and it is up-to-date and if I run the tests with `make behaviourtests` everything is okay.

But if I compile with `tools/build.sh tools/mcm_example.fac src/ model/configuration/ mcm/` it creates files in `src/` as well as in `src/gen/`. They also seem to refer to different mechanisms, so I suspect something is really wrong, somewhere.

If I delete the autogenerated files in both src and src/gen, I get an error:

```
make tools/mcm_example.fac
gfortran -o atchem2 -Jobj -Iobj src/dataStructures.f90 src/interpolationFunctions.f90 src/configFunctions.f90 src/inputFunctions.f90 src/outputFunctions.f90 src/atmosphereFunctions.f90 src/solarFunctions.f90 src/constraintFunctions.f90 src/solverFunctions.f90 src/parameterModules.f90 src/atchem2.f90 -fprofile-arcs -ftest-coverage -ffree-form -fimplicit-none -Wall -Wpedantic -fcheck=all -L/home/r/rs445/libs/cvode/lib -L/home/r/rs445/libs/openlibm-0.4.1 -Wl,-R,/usr/lib/:/home/r/rs445/libs/cvode/lib:/home/r/rs445/libs/openlibm-0.4.1 -lopenlibm -lsundials_fcvode -lsundials_cvode -lsundials_fnvecserial -lsundials_nvecserial
src/outputFunctions.f90:32: Error: Can't open included file './gen/ro2-rates.f90'
src/solverFunctions.f90:171: Error: Can't open included file './gen/mechanism-rate-declarations.f90'
```

spco:
My feeling is this must still be in your Makefile.

If I do the following:
```
rm src/gen/*.f90
./tools/build.sh tools/mcm_example.fac
./atchem2
```
then I get normal behaviour, and only those 3 files are generated in `src/gen`. This is at current `master` rather than on this branch, but I don't see any reason why this branch would change anything in that respect.

Attached is my Makefile (had to change it to a .txt just to upload it), perhaps if you `diff` against yours you will see some difference? Also please make sure you don't have a `Makefile` or `Makefile.save` in `travis/unit_tests` still hanging around.
[Makefile.txt](https://github.com/AtChem/AtChem2/files/2192571/Makefile.txt)

rs028:
@spco I figured it out. If you do

    ./tools/build.sh tools/mcm_example.fac

it puts the files in the default directory, i.e., `src/gen/`

But if you do

    tools/build.sh tools/mcm_example.fac src/ model/configuration/ mcm/

It overrides the default with `src/`. I made a small modification to `build.sh` and will make it clear in the wiki.

There is the problem of failing the macOS test, but I am tempted to merge this anyway, so that we can add it to the official release (PR #303). What do you think?

spco:
Ah yes, the interface has changed slightly. Thanks for making that read more helpfully.

Well the trouble is if we do merge it without a fix, then the tests will fail for evermore. While that's possible to do, it will mean that for future PRs we'll have to check that the failures we see are the 'right' failures.

I remember that I had to play some tricky games to get the tests to work on both systems previously - I think I had to play around tweaking some configs until I got one that was stable. So the tests are likely to fail any time we make these changes.

As another proposal, what about if we increase the relative tolerance on the tests? Change the 7.0e-6 at `/travis/run_tests.sh#L38` to 5.0e-5 and they'll all pass. 5e-5 is still not a bad tolerance to have. When we look to investigate the numerical issues, I expect I'll whack that right down to 1e-9 or something to see if I can tease out exactly where things start to diverge.

spco:
I'm about to head home, but if that passes then from my perspective I'm happy that this be merged. And I haven't been over the `CHANGELOG.md` and `README.md` changes in #303 with too fine a toothcomb, but that also looks good to merge :)

Issue #245, opened 2017-06-22
Check that tools/multiplot.py works


spco:
I'll leave that to you!

rs028:
@spco This relies on `matplotlib` and `pylab`. I can't find those on SPECTRE, so I'll have to test it on a different machine. It is an optional tool, but it would be nicer to use something that does not require additional dependencies, if at all possible. Any suggestions?

PR #267, opened 2018-04-13
Plot tools

rs028:
This is meant to address issue #245. Easier to get rid of the old script and write new ones.
Make 4 plotting scripts which create a pdf file in the output directory:
- R
- Python (with matplotlib)
- gnuplot
- Matlab/Octave


rs028:
Minor issues remain with Python and gnuplot scripts (see #269) and there is only a skeleton file for the Matlab script. Not urgent, can be merged.

Issue #265, opened 2018-03-09
Verification testing


spco:
Related to #47 .

As a step towards verifying the output to a suitable standard, I think we need a suite of tests with known solutions. Currently our tests only really cover 'did I break anything from the last commit?', but in reality we also want to have confidence that our solutions are right. Typically this is done with the 'method of manufactured solutions', and I think it would be good to have a few tests touching on that. These typically can be very uncomplicated and short to run, so they shouldn't add considerably to the tests running time. I'm thinking of a few classes of tests:

1. Constant solution. Create mechanism files that, coupled with the initial conditions, give a constant solution. e.g.
A + B -> C, rate k
A + C -> B, rate k
B + C -> A, rate k.
A(0) = B(0) = C(0) = 1/3. I think this shouldn't matter whether we fix the zenith angles etc, as the mechanism is decoupled from the temperature and lighting. I'd probably actually have 2 versions - one with light varying and one without. Both should be identical.

2. Steady-state solution. Species concentrations vary from the start, but tend toward a steady-state solution (all concentrations hit a constant value and stay there). This can be used to verify that (a) the solution converges to a steady state at all, and (b) that steady state is the correct one according to the known solution.

3. 'Manufactured solution'. We pick the RHS of the differential equations so that the solution is known. Check that the calculated solution sticks to the true solution. These can be made arbitrarily complicated, which is useful for testing the limits of what the model can do.

rs028:
Agree. Besides the current suite of tests has a lot of redundancy. You don't need a full MCM model to test a few switches.

rs028:
Marked as high priority because the new tests (see PR #268) seem to indicate some minor calculation errors. Possibily related to the numerical issues we have been worried about.

spco:
Just putting this here as it seems the most appropriate place: when looking at numerical stability, difference in the versions of BLAS/LAPACK should be considered as a possible source of issues. More rigorous unit testing of our own routines will help narrow this all down.

rs028:
The accuracy of the solver (i.e., the default values in `solver.parameters`) needs to be investigated. See notes in (https://github.com/AtChem/AtChem2/issues/384#issuecomment-494454524) for more information.

rs028:
Additional notes about accuracy and stability of the solver:
- need to rationalize and expand the test suite - baseline numerical tests are needed (see #47, #326)
- need to upgrade CVODE (see #257)
[...]

PR #268, opened 2018-04-18
New tests

rs028:
I started assembling some simple models that can be solved analytically. This is to address issues #47 and #265. At the moment the tests are not executed by travis or `make test`.


spco:
When this comes to merging, we'll need to redo the new test outputs, as #280 fixed a bug that affects the output.

rs028:
We will have to decide what to do regarding the testing regime. At the moment is, I think, overly complicated/redundant. I'd like it to be simpler and more systematic. In my mind this branch was supposed to be the starting point for a new set of tests to replace the current ones (sometimes it's easier to start over from scratch, then trying to fix something but maybe it is just me). Any thoughts?

spco:
I think some of the existing tests should certainly be dropped, though it is probably worth having a overlap period where we introduce the new ones but keep the others in place for a while. Does that sound reasonable? We can use codecov to check that our new tests cover all the code lines that the old ones do before deprecating the old ones. I agree that several of the existing tests are overkill and overlap too much.

spco:
We also need to manufacture a few minimal tests that trigger the numerical differences between Linux and Mac, to help diagnose the cause. If we can strip out as much as possible and still see the issue then that helps.

rs028:
Yes I agree. I think it's better to have more, but simpler/shorter tests, rather then the current system which runs complete models.
I am also wondering if it would be good to have a simple text file in `travis/` which each test and what it does, so it is easier to keep track of them?

spco:
Such a text file might be useful - not sure how to best implement that. In the meantime, can you tell me what the tests in this PR do? What are their analytic solutions?

rs028:
I was thinking something really simple as:

    test1 : two reactions mechanism to test correct use of switches A, B , C
    test2 : two reaction mechanism to test contraining of chemical species
    test3 : three reaction mechanism to test contraining of environment variables

I will email you about the tests in this PR.

Issue #272, opened 2018-06-06
Simplify photolysis rates read


spco:
The reading in of photolysis rate is a bit of a mess currently:
```
readPhotoRates()
 `-> readPhotolysisConstants()
      `-> readPhotolysisRates()
```
is a nightmare structure for knowing which calls which, and this is the only time they're called. At the least, these need renaming, but I think it would be better to combine that with a rethink of the structure - these shouldn't really be nested.  Better to simplify the logic of the whole thing.

spco:
The fix in #273 adds a slight annoyance in that the filename of `photolysisConstants.config` is used in two places. This should be neatened up before this issue is closed.

spco:
@rs028 A couple of questions on this (not urgent to answer): I think the current logic is that there are two types of photolysis possible at any one time: we have either (a) some rates are fixed constants (from `photolysisConstants.config`), and some are (varying with time) constrained (`constrainedPhotoRates.config`) ; or (b) some rates are calculated (`constrainedPhotoRates.config`), and some are (varying with time) constrained (`photolysisRates.config`). 
`nrOfPhotoRates` hold how many are either constant or calculated, and `numConPhotoRates` how many are constrained.

(1) It's possible to supply the same species in both constant and constrained, or calculated and constrained. In this case, the calculated or constant gets overwritten by the constrained. I guess to be a bit more efficient we could check whether there's any overlap, and remove the overlapping ones from the calculated or constant group - the constrained rate will always be applied anyway, so why waste time calculating? What do you think? And is this the right order of prioritisation, or should e.g. the calculated rate always trump the others?

(2) Do we want this to be exclusive (only constants or calculated allowed, never a mix) or would it be useful to have some calculated and others fixed constants? At present that can be 'hacked' by supplying a constant value in the constraints file, but I think supplying both a constants and a rates file should be allowed, as well as the constraints file. Thoughts?

rs028:
@spco I haven't really looked at how the photolysis rates are implemented so I cannot comment on everything. That being said this is how I think it should work.

case 1: all photolysis rates are calculated. Using the info in `photolysisRates.config` and the MCM parametrization (see http://mcm.leeds.ac.uk/MCMv3.3.1/parameters/photolysis_param.htt)

case 2: all photolysis rates are constrained.

case 3: some photolysis rates are constrained, those that are not constrained are calculated.

[...]

PR #273, opened 2018-06-06
Remove nested nature of readPhotolysisConstants/readPhotolysisRates

spco:
Only one of them needs to be called, based upon the existence of param_dir/photolysisConstants.config, so only call one.

This partially addresses #272, though more remains to be done.


spco:
@rs028 I'll leave this for you to review. While there's more to do on this topic, this at least stops the slight insanity of `readPhotlysisConstants()` calling `readPhotolysisRates()` if it fails, making the flow clearer. Could you Squash and merge if you're happy?

PR #278, opened 2018-06-08
Alter .travis.yml and Makefile to set up CodeCov.

spco:


spco:
@rs028: this looks really useful for testing purposes - it's very easy to see the lines that never execute in our tests so far, so that should guide our future testing plans! I think the big comment by codecov-io above is only because it had no previous to compare against - I'm not sure what it will normally show.

rs028:
Very interesting!

Issue #271, opened 2018-06-06
Defensive programming


spco:
This is an issue to catch any areas we could improve the defensive programming:
 - [ ] `getConstrainedQuantAtT` should check that `1 <= ind <= size( x, 1 )`
 - [ ] `getConstrainedQuantAtT`: Need to check that `dataNumberOfPoints+1` doesn't go out of bounds
 - [ ] `readSpeciesConstraints` l942 should abort if the `if (id/=0)` is false, like in l861

PR #283, opened 2018-06-12
Defensive programming fixes

spco:
Check supplied index is within bounds, fix a bug which would go outside of bounds, and output an error if a supplied species is not a member of the problem.


spco:
Fixed #271

PR #285, opened 2018-06-13
Remove makefile.local usage - users should just edit Makefile directly.

spco:
This needs the documentation to be changed. This is a bit tricky, because the logic in Makefile is a bit complicated, for the sake of Travis. I've made an attempt to simplify that.


spco:
@rs028 This is ready to merge from my perspective. What do you think? The user now just replaces the lines `CVODELIB = ...` and `OPENLIBMDIR = ...` lines with the appropriate locations.

The slight downside to this setup (from the developer's view) is that, because `Makefile` is a tracked file, if we set these lines for our local development then we need to remember not to add these changes into new commits. That just requires discipline, and should be picked up by Travis anyway, because the executable won't build. Is this all ok?

If so, https://github.com/AtChem/AtChem2/wiki/1.-Installation needs alteration to document the new easier setup.

rs028:
Looks okay. But what about BLAS and LAPACK paths?
Regarding the tracking would it be easier to put Makefile in .gitignore and have an example Makefile in tools?

spco:
I could be wrong, but I can't find where the `BLASDIR` and `LAPACKDIR` are ever used - I think these must have been removed at some point, but the documentation not updated? Slightly strangely, I can't even find them in the history anywhere, so who knows.

rs028:
So they are only needed to install CVODE? Maybe CVODE retains the information of where the two libraries are located?

spco:
Looking at the `atchem2` executable, it doesn't link against lapack or blas at all, so it doesn't need them except through CVODE. CVODE should certainly retain where they are located when installed. CVODE installation via `tools/install/install_cvode.sh` doesn't use these variables either (it uses hardcoded default paths for `LAPACK_LIBS`). So it does seem we don't need these at all. So long as CVODE is installed correctly, AtChem doesn't care.

rs028:
All right then. But what about `-lblas -llapack` flags in the Makefile?

spco:
That's a good point - I'll have to check that tomorrow.

On the issue of tracking, I'm not sure. Moving it elsewhere requires a new user to copy the file to the correct directory before editing it. Is that reasonable to expect of a new user, or not? I guess it's a case of either inconveniencing the user or the developer! And it would only need to be done the once when first installing/ after cloning.

spco:
Looks like we don't need those flags either, so I've removed them.

rs028:
Okay good. This means, I suppose, that BLAS and LAPACK are not dependencies, but are only needed to install (and eventually upgrade) CVODE, so we'll have to update the documentation accordingly.

Re the `Makefile`, I think it is reasonable to expect the user to be able to copy the skeleton file from `tools/`. It is not such a big step and the user will have to open the file and modify it anyways.

spco:
Ok, so in summary, this PR removes the `makefile.local` mechanism, removes explicit dependence on LAPACK and BLAS (only dependent via CVODE), and moves the `Makefile` to `tools/` and adds `Makefile` to `.gitignore`. Users now need to `cp tools/Makefile Makefile`, but it gives the benefit that developers won't be annoyed by `Makefile`.

If you are happy @rs028, I will merge this. Please let me know. I will then also change the wiki to reflect these changes.

rs028:
@spco yes I happy to merge this.

PR #286, opened 2018-06-14
Upgrade cvode

spco:


spco:
Should this be a breaking change, or perhaps use preprocessor directives? It might be useful to change the wiki to tell users to install cvode into a cvode2/3 directory.

spco:
The latter is certainly the less messy and easiest from a development point of view. I'd suggest we go with that, but it might sway us into having to make this a 2.0 release, perhaps? It's quite a big compatibility change.

rs028:
You mean this PR was not meant to be merged with #303? I am not exceedingly worried about the versioning, tbh.

I think it is okay to overwrite CVODE with the new version.  We just have to document the behaviour. Users who want to use a previous version of either Atchem or CVODE are probably also skilled enough to install them by themselves.

spco:
Upon reflection, I think this should probably go into v1.1 actually, so it can go before #303. I agree with your rationale for not caring about backwards compatibility at this stage.

spco:
So unfortunately this seems to break the carefully-balanced behaviour tests in quite a few places. Essentially, the solver change means we get differences in 8 linux tests (and 1 of the MacOS for good measure too). I'm not sure that I want to go through the whole rigmarole of tweaking the tests to pass - that seems rather a backwards way to do things. Perhaps it's better that I do some more unit testing and then focus on getting some of the exact-solution tests out of the door?

rs028:
Yes, I think that would make more sense. I don't mind postponing the upgrade to CVODE. I'd rather have solid results.

rs028:
I am closing this, as it is superseded by PR #545.

Issue #270, opened 2018-05-18
Various minor issues


rs028:
- [ ] check that initial and and final model time are within the initial and final constraint timestamps before starting interpolation

spco:
`Makefile.local` is solved by #285

spco:
`constrainedFixedSpecies.config` is for constant-value species. Its name will become `speciesConstants.config` or `speciesFixed.config` in the Great Reorganisation.

spco:
If you mean `mechanism-rate-declarations.f90`, then yes, it's used in `mechanism_rates()` in `solverFunctions.f90`.

spco:
On 1., the current behaviour in `getConstrainedQuantAtT()` is this: if the required time is within the time covered by the constraint data, then it interpolates. Otherwise, it returns the value of the _last_ constraint data.

This certainly needs fixing in at least one case: if the required time is _before_ the first constraint data time, then the _last_ constraint data value is still returned. This is obviously not what is wanted.

I've opened #294 for this - see discussion there.

spco:
On 8., both photolysis rates and environment variables are only evaluated when needed - each is interpolated individually, only when constrained. This happens each time `mechanism_rates()` is called from `FCVFUN()`, so the number of times that happens is controlled by CVODE as it completes the integration.

Species interpolation is very similar - it is called once for each of the constrained species in `FCVFUN()`, plus once when setting the initial conditions of each of the constrained species.

At this time I don't see a way to reduce the cost of this - once I've profiled more extensively, that will shine a light on how expensive this all is, and whether it's worth looking harder at.

rs028:
@spco I added this to the Milestone for version 1.1 because there are only a few points left and some of them are just questions to improve the documentation. If some require more time than you have at the moment, they can be moved to their own issue.

spco:
On number 11 above - it shouldn't matter if the same species is mentioned twice in the same file of either `speciesConstant.config`, `speciesConstrained.config`, `photolysisConstrained.config`. It's inefficient, as it'll all be read in and stored and processed, but it shouldn't matter to the result. But it  looks like we don't check that there's no conflict between species defined as constant or variably constrained? I'll make this a separate issue.

spco:
On 3. above, currently we process RH, and then H2O. RH can be supplied as a constrained, or fixed value. H2O is then either calculated from RH (via `convertRHtoH2O()`), or set as constrained, set as fixed, or given a default value. There are 2 issues with this setup currently:
[...]

PR #287, opened 2018-06-20
Initial work on reorganising the configuration and output directories.

spco:
This would need some more changes to be consistent in e.g. the README, but this moves to the following directory structure:
```
atchem2
|- model
|    |- environmentConstraints
|    |- modelConfiguration
|    `- speciesConstraints
|- obj
|- output
|    |- instantaneousRates
|    `- all other .output files
|- src
|    |- all .f90 files
|    `- auto-generated files (mechanism-rate-coefficients.f90, mechanism-rate-definiitons.f90 and ro2rates.f90)
|- tools
|    |- install
|    |- plot
|    `- other tool files
`- travis
```
What do you think of this as a first draft? I think `tools` might still want some further work: perhaps all the python scrits for the MCM could go into their own `python` directory or similar. 

And `RO2listv3.3.1` should perhaps be moved elsewhere - it's not a tool, perhaps it would be better to go in `model/modelConfiguration`? It is then at least in the same directory as `photolysisRates.config`, which partially addresses one of the notes in #270 about easier swapping in and out of MCM versions.


rs028:
Good first draft. I am thinking about a possibile workflow.
Say I have a project and all the models for the project are in the Atchem directory:

    atchem2
    |- model1
    |- model2
    |- model3
    |- src
    |.....

So I setup and run model1, then I modify the mechanism, rebuild and run model2, then I change the configuration and without rebuilding I run model3, etc....
Being able to do this is the main rationale behind the reorganization I think, but it cannot create conflicts between each model (output, configuration or executable)
On the other hand if everything is in each model's directory, we might as well keep a flat structure and copy the whole Atchem directory over for each model. Does this make sense?

spco:
I'm not sure I understand, so apologies if the below doesn't make sense.

If we want to move to a function-parser based system (I'm taking a little look at that this afternoon), then we won't need to recompile, so should ideally only have one executable (we need to talk more about whether/how that's possible if you also want to be able to add new environment variables).

I see the workflow being that you have `model1`, `model2` as above, but they don't need to live under `atchem2` directory - they can live absolutely anywhere. Let's say they live in the user's `$HOME` directory, for argument's sake. Then we run `atchem2` with the correct command line arguments to point to those directories as input and output directories.

The `model` and `output` directories inside `atchem2` are just there as default directories, and show the user an example of how things should be structured. The typical workflow should ignore those and use user-defined directories via the command line arguments.

This also fits with the aim to begin to separate the executable from the user-defined input files, which is good practise. For example in an HPC setting, you could have a pre-compiled `atchem2` installed as a module on ALICE2, which would live in `/cm/shared/apps/atchem2`. This is not writeable by users, so the user can't modify files in `/cm/shared/apps/atchem2/model/modelConfiguration` etc. Instead, the user has `model1` in `$HOME` as above, and points `atchem2` to those directories when running it.

If that answer was based on misunderstanding your previous post, please correct me!

rs028:
No you are right. I was not thinking about the function parser.
But then shouldn't `output` be a subdir of `model` so that the config files and the output stay together?

spco:
Ok, yes, that would probably make sense. That looks like

```
atchem2
|- model
|    |- environmentConstraints
|    |- modelConfiguration
|    |- speciesConstraints
|    `- output
|         |- instantaneousRates
|         `- all other .output files
|- obj
|- src
|    |- all .f90 files
|    `- auto-generated files (mechanism-rate-coefficients.f90, mechanism-rate-definiitons.f90 and ro2rates.f90)
|- tools
|    |- install
|    |- plot
|    `- other tool files
`- travis
```

A related note (because it relies on the structure of the directories):

I'm thinking that the new command line interface should take as input a `base` input and `modelconfig`, `output`, 'envcon' etc inputs. If `base` is supplied, then it should point to the `model` directory, with all the other directories inside as below. If `base` is not supplied, then it defaults to the directory `atchem2` is in. But the other inputs can also be supplied, and they overrule.

So e.g. `./atchem2 --base=$HOME/model1 --envcon=$HOME/model2/environmentConstraints` would use `modelConfiguration`, `speciesConstraints`, and `output` from `$HOME/model1` but `$HOME/model2/environmentConstraints` for the environmental constraints.

Normal usage will be just to supply `--base`, but we have the flexibility to mix in other bits if the user wants.

(I think I'd want to change the word `base` to something else)

rs028:
Okay, how about this:

```
atchem2
|- model
|    |- environmentConstraints
|    |- modelConfiguration
|    |- speciesConstraints
|    |- modelOutput
|         |- instantaneousRates
|         |- error and diagnostic files (mainSolverParameters.output, erros.output, jacobian.output)
|         `- all other .output files
|- obj
|- src
|    |- all .f90 files
| - mecha
|    ` - mechanism.fac
|    ` - mcm files (RO2listv3.3.1, photolysisRates.config)
|    `- auto-generated files (mechanism-rate-coefficients.f90, mechanism-rate-definitions.f90, ro2rates.f90)
|- tools
|    |- install (include mcm_example.fac, Makefile)
|    |- plot
|    |- build
|    `- other tool files (fix_*.py)
`- travis
```

The extra directory `mecha` includes all the MCM relevant data. Part of it I guess could go in modelConfiguration but we still need a place for RO2List. I am not sure about the autogenerated files, but I suppose they willo disappear with the function parser.

Yes to the command line interface.

spco:
That looks fine (I just need to check that having the auto-generated files in `mecha` doesn't cause us problems when compiling - it shouldn't do, and as you say this should be reduced by the function parser later).

rs028:
If it takes more than 5 minutes to fix the autogenerated files and they will soon disappear, than we can just leave them there.

Another thought about the model directory.  Would it make sense to group the input files? Would it make easier to deal with the photolysis rates as discussed in #272?

```
|- model
|    |- configuration
|    |- constraints
|         |- species
|         |- environment
|         |- photolysis
|    |- output
|         |- instantaneousRates
|         |- error and diagnostic files (mainSolverParameters.output, erros.output, jacobian.output)
|         `- all other .output files
```

 I don't want to create an overcomplicated/rigid structure and I tend to prefer flatter directory structures, so this is another option.

```
|- model
|    |- configuration
|    |- species
|    |- environment
|    |- photolysis
|    |- instantaneousRates
|    |- error and diagnostic files (mainSolverParameters.output, erros.output, jacobian.output)
|    `- all other .output files
```

(Also renamed some directories. I am partial to simple names for files and directories, but I am happy to overrule my preferences to follow best practices).

rs028:
@spco have we converged on a new directory structure yet? I got lost between the different proposals in here ;)

spco:
I'm not sure that we quite have yet. I agree that putting `output` inside `model` would help easy recognition of the pairing of inputs to outputs. I think the first structure [here](https://github.com/AtChem/AtChem2/pull/287#issuecomment-398767725) makes the most sense to me - it identifies that the `environment` and `species` directories are about constraints. Having the `constraints/photolysis` subdirectory also makes a lot of sense - it's anachronistic that the constrained photolysis rates live in `environmentConstraints` currently. I don't know whether it would be best to also move `constrainedPhotoRates.config`, and possibly also `photolysisRates.config` and `photolysisConstants.config` to `constraints/photolysis`? Or leave them in `model/configuration`.

In short, here are a couple of proposals: (a) e.g.
```
model
   |- configuration
   |    |- concentrationOutput.config
   |    |- constrainedFixedSpecies.config
   |    |- constrainedPhotoRates.config
   |    |- constrainedSpecies.config
   |    |- environmentVariables.config
   |    |- initialConcentrations.config
   |    |- lossRatesOutput.config
   |    |- model.parameters
   |    |- photolysisConstants.config
   |    |- photolysisRates.config
   |    |- productionRatesOutput.config
   |    `- solver.parameters
   |- constraints
   |    |- species
   |    |     |- CH4
   |    |     `- O3
   |    |- environment
   |    |     |- PRESS
   |    |     `- TEMP
   |    `- photolysis
   |          |- J3
   |          `- J12
   `- output
        |- instantaneousRates
        |- error and diagnostic files (mainSolverParameters.output, erros.output, jacobian.output)
        `- all other .output files
```
(with a couple of changes to filenames as well - I dislike the fact we have `photolysisRates.config` and `photolysisConstants.config`, but `constrainedPhotoRates.config`. Or (b)
```
model
   |- configuration
   |    |- concentrationOutput.config
   |    |- lossRatesOutput.config
   |    |- productionRatesOutput.config
   |    |- model.parameters
   |    `- solver.parameters
   |- constraints
   |    |- species
   |    |     |- constrainedFixedSpecies.config
   |    |     |- constrainedSpecies.config
   |    |     |- initialConcentrations.config
   |    |     |- CH4
   |    |     `- O3
   |    |- environment
   |    |     |- environmentVariables.config
   |    |     |- PRESS
   |    |     `- TEMP
   |    `- photolysis
   |          |- constrainedPhotoRates.config
   |          |- photolysisConstants.config
   |          |- photolysisRates.config
   |          |- J3
   |          `- J12
   `- output
        |- instantaneousRates
        |- error and diagnostic files (mainSolverParameters.output, erros.output, jacobian.output)
        `- all other .output files
```
This has the benefit that it really does group things together, but it is less obvious to know where config files are, as they're split up across a number of directories.

What do you think of either option?

rs028:
I think the first one is better. I prefer to have all configuration files in one directory.

`photolysisRates.config` is not really a config file and should be moved (see PR #290); and I agree `constrainedPhotoRates.config` and `photolysisConstants.config` should be renamed: how about `constrainedPhotolysis.config` and `constantPhotolysis.config`?

One other thought: is it necessary to separate the type of constraints in `species`, `environment` and `photolysis`? or they could be all in the same directory? From the point of view of the user I guess it is the same, but it may be easier for the code if they are separated?

spco:
I'm happy to go with the first one - as you say, it's nice to have the configs together. I'll leave the issue of `photolysisRates.config` out of the PR, as it's handled in #290 as you say. 

Could I suggest that we use the alphabetic ordering to group the config files a bit more, and use that to guide our file naming scheme? E.g.
```
concentrationOutput.config
constrainedFixedSpecies.config
constrainedPhotoRates.config
constrainedSpecies.config
environmentVariables.config
initialConcentrations.config
lossRatesOutput.config
model.parameters
photolysisConstants.config
photolysisRates.config
productionRatesOutput.config
solver.parameters
```
becomes
```
environmentVariables.config
model.parameters
outputConcentration.config
outputLossRates.config
outputProductionRates.config
photolysisConstants.config
photolysisRates.config
photoRatesConstrained.config
solver.parameters
speciesConstrained.config
speciesFixed.config
speciesInitialConcentrations.config
```
which allows much easier at-a-glance understanding of whether any photolysis rates are constants or constrained, same for species. And helps you find the files controlling output, even if you can't remember the exact name of the file that controls that type of output.

spco:
And on the issue of whether to separate the constraints into directories, I think it's actually for the user's sake more than the code. If we merge them into one directory, the user may have to look through lots of constrained species to see whether they've got any constrained photolysis/environment bits etc. The user can immediately compare the contents of `environmentVariables.config` with the contents of `environmentConstraints` directory to check that all that are `CONSTRAINED` have a matching file.

rs028:
Okay so to recap:

```
atchem2
|- model
|    |- configuration
|    |- constraints
|           |- species
|           |- environment
|           |- photolysis
|    |- output
|         |- instantaneousRates
|         |- error and diagnostic files (initialConditionsSetting.output, mainSolverParameters.output, erros.output, jacobian.output, finalModelState.output)
|         `- all other .output files
|- obj
|- src
|    |- all .f90 files
| - mecha
|    ` - mechanism.fac
|    ` - mcm files: RO2listv3.3.1, photolysisRates.config (renamed)
|    `- auto-generated files: mechanism-rate-coefficients.f90, mechanism-rate-definitions.f90, ro2rates.f90 (can provisionally remain where they are now if too difficult to move)
|- tools
|    |- install (include mcm_example.fac used for testing and Makefile example)
|    |- plot
|    |- build
|    `- other tool files (fix_*.py)
`- travis
```

with configuration files renamed to be alphabetical, as you suggest. On that matter, the output files could also be renamed in a more logical way.

spco:
Yes, that looks good to me. I'll proceed with that unless you shout otherwise!

rs028:
@spco in this structure all directories have full names, not abbreviations. So the new `mecha` should be called `mechanism`.

rs028:
It looks good. 

Re the `mechanism` folder, I am a bit undecided. My initial thought was to keep all the MCM related stuff in one folder. but that definition really only applies to `photolysisRates.config` and `RO2listv3.3.1`. The mechanism is provided by the user and I can see the possibility of running different mechanisms with the same configuration settings (in a kind of sensitivity study), so it would maybe make more sense. Maybe it would be better to have `mechanism.fac` in `model/configuration/` and the autogenerated files in `obj/` or `src/`?

Regarding the photolysis rates, what I have done in the past is to use one of the "empty slots" to add a constraints. I have not added the l, m, n parameters, because finding those parameters would require a fitting code that I don't have and it is something done during development of the MCM. So I think it is fine to keep that file as static.

spco:
Ah OK, so the slots are only used for constraints, not new parametrized rates. Thanks for clearing that up.

How about we place the autogenerated files into a new directory, `src/generated` or `src/gen`? Eventually those files would hopefully disappear anyway with the function parser, so the directory would then be removed. It just keeps it obvious which files are dependent.

And perhaps moving the `.fac` to `model/configuration` is a good point - it can in general be swapped in and out, independent of the `photolysisRates.config` and `RO2listv3.3.1`. Then `mechanism/` only contains those two files, and can easily be swapped out - e.g. have two directories called `mechanism3.1` and `mechanism3.3.1`, and either can have its contents copied into `mechanism/`. Or use symlinks etc. Either would work, but that's another PR.

Are you happy if I do as we're discussing?

rs028:
Yes, good idea.So let's move the `.fac` file to `model/configuration` and the autogenerated files to a directory in `src/`. In fact we could get rid of `mechanism/` altogether and move the MCM files in `src/mcm` or something like that. What do you think?

spco:
`src/` by convention contains the source code, rather than data. Normal users typically don't need to look in `src/`. I think keeping it as `mechanism/` would make sure it's visible to users as something they can alter, if they wish?

rs028:
Okay, then.

rs028:
In this case `mechanism/` should probably be renamed as `mcm/` (or something like that) to avoid confusion that the `.fac` file does not go in there

Issue #289, opened 2018-06-21
Rename MCM files to include version number and be descriptive


spco:
From https://github.com/AtChem/AtChem2/issues/272#issuecomment-399058957

PR #290, opened 2018-06-22
Update MCM related files

### LINKED ISSUE #288: photolysisRates.config doesn't match MCM version
http://mcm.leeds.ac.uk/MCMv3.3.1/parameters/photolysis_param.htt doesn't exactly match `photolysisRates.config`: in particular, the latter has a <57> which the former lacks. I haven't checked whether the rest is identical. What should be done with this?

rs028:
Fixes #288 and #289. These files are in a provisional folder -> see PR #287.


spco:
Would you be able to update this to the latest directory setup?

rs028:
I think I have done it, but I am not sure, my `git status` looks messy :(

rs028:
@spco I always get confused when I need to sync the repos. I think this is all right although there is a ghost commit (`fa86ff9`) but it should not matter. Can I squash and merge? (or rebase and merge, not clear about the diff in this case)

spco:
I think a squash and merge would be best in this case, as the second commit just undoes the first.

Before that though, what is the mechanism we want to use for how to swap these files in and out by version? Perhaps for now it is best just to hard-code in the one or two places we have - we can come back and make it pretty later, and it's what we already do for the RO2listv3.3.1 file. In that case, could you change the references in the code and tool-scripts to `photolysisRates.config` and `RO2listv3.3.1` to point to the new files? Might be nice to have that in this PR?

rs028:
Ok. This PR renames the files as `photolysis-rates_v3.3.1` and `peroxy-radicals_v3.3.1`  so I also deleted `photolysisRates.config` and `RO2listv3.3.1` which are duplicate.

tools/mech_converter.py
```diff
@@ -191,7 +191,7 @@ def convert(input_file, output_dir, mc_dir, mcm_dir):
 
     # check RO2s are in RO2 list
     # Read in RO2 list, and strip off newlines
-    with open(os.path.join(mcm_dir, 'RO2listv3.3.1')) as RO2List_file:
+    with open(os.path.join(mcm_dir, 'photolysis-rates_v3.3.1')) as RO2List_file:
```
spco: I think this should be peroxy-radicals?

tools/mech_converter.py
```diff
@@ -191,7 +191,7 @@ def convert(input_file, output_dir, mc_dir, mcm_dir):
 
     # check RO2s are in RO2 list
     # Read in RO2 list, and strip off newlines
-    with open(os.path.join(mcm_dir, 'RO2listv3.3.1')) as RO2List_file:
```
rs028: should it be : `with open(os.path.join(output_dir, 'gen/mechanism-rate-coefficients.f90')`

spco: No, because in `setup_atchem2.py` line 31 sets `output_dir` to `src/gen` already - that is abstracted away from this script.

rs028:
Then something else is wrong, because it compiles but does not pass the tests. And there are a bunch of `*.gcda` and `*.gcno` in the main directory

spco:
The `*.gcda/gcno` files are to do with codecov - they are ignored by git.

It looks to me like it's not actually compiling the first test on Travis - the `atchem2` binary doesn't exist when running the first test.

For starters, each of the tests currently has a `photolysisRates.config` and `RO2listv3.3.1` in its own `mcm` directory, which all need changing to the new name? I expect `build.sh` is failing for each test (silently, because we mute that script for brevity) when it can't find those files.

rs028:
All right, that was it, but it still fails 8 tests. Did you say some of the tests used an old version of the photolysis rates?

RE the `*.gcda/gcno` files are these going to be always produced at compilation? Should they be moved to obj, with the other compilation files?

spco:
Yes, spec_no_env_yes1, spec_no_env_yes2, spec_yes_env_no, spec_yes_env_no_with_jfac, spec_yes_env_no_with_jfac_fail1, spec_yes_env_no_with_jfac_fixed, spec_yes_env_no_with_photo, spec_yes_env_yes, spec_yes_plus_fixed_env_no are all different from `mcm/photolysisRates.config`.

https://gcc.gnu.org/onlinedocs/gcc/Gcov-Data-Files.html suggests the `.gcda` files can be moved elsewhere with a compiler flag, but it doesn't seem clear whether the same is true of the `.gcno` files.

spco:
Yes, I think that's going to have to be the case, unless we do as below. Hopefully that doesn't break under the numerical issues we have.

Suggestion: we don't hardcode the version numbers, instead only making reference to `peroxy-radicals` and `photolysis-rates` in the source code. We can then just rename all the test files to exactly that. In the base `mcm`, add a symlink `photolysis-rates -> photolysis-rates_v3.3.1` and the same for the other file. We can then change the version number by changing the symlinks to point to the other version. What do you think? I think a negative to this is that that may need to be set up by the user - may not be able to encode that within the git repository?

rs028:
This should be fine now (I'm leaving for the day, so feel free to merge it if you are happy with it). I guess the `mcm/` folders in the tests are pointless now, but the tests are planned to be substituted anyways so we can remove those at that point.

For the moment I would leave the version number hardcoded. I think most people would use the latest version of the MCM anyways. The symlink is a possibile solution, another is to add a switch in a configuration file. I will open an issue with this but tag it as lower priority.

PR #291, opened 2018-06-26
Photolysis reconfig

spco:
While not quite production ready (there are a few bits I want to tweak), this now passes all current tests, except for the fact that it modifies the screen output of each test (because the execution order changes, so the reference screen output needs to be updated) and a small change to the way photolysis rates are output in `photolysisRates.output`. Rather than all photolysis rates being output in the order they come in `photolysisRates,config`, we have them ordered as firstly the unconstrained rates (those that are calculated) and then the constrained rates.

Is this acceptable @rs028? It would be a bit of a pain to reorder them.

### DIFF
--- a/src/outputFunctions.f90
+++ b/src/outputFunctions.f90
@@ -125,7 +125,8 @@ contains
     use types_mod
     use photolysis_rates_mod, only : ck, j, constantPhotoNames, numConstantPhotoRates, &
                                      constrainedPhotoNames, numConstrainedPhotoRates, &
-                                     unconstrainedPhotoNames, numUnconstrainedPhotoRates
+                                     unconstrainedPhotoNames, numUnconstrainedPhotoRates, PR_type, &
+                                     constrainedPhotoRatesNumbers
     implicit none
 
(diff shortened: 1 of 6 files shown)


src/outputFunctions.f90
```diff
@@ -123,19 +123,48 @@ end subroutine outputPhotoRateCalcParameters
...
 
     real(kind=DP), intent(in) :: t
     integer(kind=NPI) :: i
     logical :: firstTime = .true.
 
-    if ( firstTime .eqv. .true. ) then
-      write (58, '(100A15) ') 't', (trim( photoRateNames(i) ), i = 1, numPhotoRates)
-      firstTime = .false.
+    ! Output constant photolysis rates if any.
+    ! Otherrwise, output constrained (if any), then unconstrained (if any).
```
rs028: Typo here: "Otherrwise"

rs028:
Yes, it's fine. I don't think it matters in which order they are output.

src/outputFunctions.f90
spco: 👍

PR #292, opened 2018-07-02
Use FRUIT for unit testing

spco:


spco:
@rs028  We now have unit testing :smile: . I'll add something to the wiki about it, and then we need to get on and add more unit tests.

I've identified one or two changes already to make, but I won't be able to do many - I've been called up to move onto another project for now, so once I make a few changes I've spotted now, I'm probably going to go quiet for a few months.

rs028:
Awesome! Good job!

Issue #10, opened 2017-03-21
Implement version markup


rs028:
- remove svnversion from Makefile and atchem.f
- delete svn_version.f

Question:  is a version markup needed? if so can we swap the svn generated code with a git generated code?

spco:
I'm currently pondering this one - a version markup can be useful, especially when it comes to reproducibility of results. However, it's impossible to hardcode the git commit hash into the commit, as the hash is a function of the contents of the commit! So I'm not sure what to replace this with.

For now, I've ripped it out and not replaced it. We can come back to this later on once the dust settles.

rs028:
Agreed. It would be useful, but can wait.
Maybe a date stamp could be easier to implement/more meaningful than a version number?

spco:
Note: svn printing mechanism was removed in 421f5f101.

rs028:
@spco this may be relevant: https://help.github.com/articles/about-releases/

spco:
Yes, I'm definitely planning to make use of that functionality. I'm still not sure how to add in a more accurate version number than just hard-coding the latest major or minor release.

rs028:
Okay. It's low priority anyways

spco:
Closed by #293. We must just remember to update this each time we release a new version, or start dev work on the next.

PR #293, opened 2018-07-03
Print version at top of output

spco:
This is the simplest way to address #10. It just requires us to remember to update this to 'v1.1' as the last thing before we release version 1.1. Thoughts?

### DIFF
--- a/src/atchem2.f90
+++ b/src/atchem2.f90
@@ -113,7 +113,7 @@ PROGRAM ATCHEM2
   rout(:) = -1.0_DP
   rpar(:) = -1.0_DP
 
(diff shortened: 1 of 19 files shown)


spco:
Sorted. Are you happy if I merge this?

rs028:
This is probably the easiest way. It occurs to me whether the whole printout should be redirected to a log file. It gets pretty big for a long model run. But maybe it's better to do it in other PR.

spco:
I think I'd prefer to leave that to the user to direct with `> atchem2.out` or similar, but I'm not too fussy. That could also be a command line option -`-o atchem2.out` etc.

rs028:
All right. Let's merge this.

PR #300, opened 2018-07-10
Move unittest build to base Makefile.

spco:
Need to remove the travis/tools/Makefile too


spco:
This improves on #292.

spco:
I've now updated the wiki to reflect the use of unit tests, and instructions on how to add new tests. Any questions, do ask - it is a bit tricky if you don't get the exact right syntax.

spco:
And of course, as I've updated `tools/Makefile`, you'll have to manually copy that to `Makefile` (keeping your existing machine-specific edits) to have access to the new `make tests` (including unit tests) and the `make unittests` and `make behaviourtests` targets.

Issue #299, opened 2018-07-10
Replace all 1.0d+8 -style constants with 1.0e8_DP style


spco:
#298 highlighted the need for this, and it leads to noticeable differences. I tested against calculations in other languages, which confirmed to me that the `_DP` style is higher-precision.

Ideally, all instances where this style is used should be replaced. It's not hard to find them with a simple regex.  `[0-9]*\.[0-9]*d[+-]*[0-9]+` This would be more useful to look for than the 'single-precision constants' the perl script looks for in `tools/Makefile`. This shows up about 8 outstanding instances.

I expect it might also be useful to change the `mc_converter` script to change all the constants from the `.fac` files to this format. Also worth looking at `photolysis_rates_v*`.

spco:
Fixed by #302

PR #302, opened 2018-07-12
Address non-generated D+08-style constants

spco:
Change remaining non-generated constants to use e+08_DP instead of D+08 form. Also change generated line for ro2.

Addresses #299, but not completely - need to look into whether the generated files should use this form too.


spco:
This evidently makes no large difference on Travis or my Mac, but that's a good thing to have confirmed :smile: It's conceivable that these parts would have previously failed unit test though, if we had one covering them.

PR #306, opened 2018-07-16
Remove unnecessary dependency in tools/setup_atchem.py

spco:


spco:
Hoping to sneak this in before you complete the release!

PR #307, opened 2018-07-30
Further unit tests for atmosphere functions

spco:
Add tests for convertRHtoH2O(), and use parts-per-unit rather than parts-per-million to reduce numerical effects. Add another test for calcAtmosphere().

### DIFF
--- a/src/atmosphereFunctions.f90
+++ b/src/atmosphereFunctions.f90
@@ -63,7 +63,7 @@ contains
     implicit none
 
     real(kind=DP), intent(in) :: rh, temp, press
-    real(kind=DP) :: h2o, h2o_ppm, temp_c, wvp
+    real(kind=DP) :: h2o, h2o_ppu, temp_c, wvp
 
     ! Use Eq.6 to calculate the water vapour saturation pressure and
     ! Eq.1 to calculate the water vapour pressure from relative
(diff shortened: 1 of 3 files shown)


src/atmosphereFunctions.f90
```diff
@@ -72,11 +72,14 @@ pure function convertRHtoH2O( rh, temp, press ) result ( h2o )
...
-    ! Eq.18 (see Vaisala paper).
-    h2o_ppm = 1.0e+06_DP * wvp / (press - wvp)
-
-    ! convert ppm to molecule cm-3
-    h2o = h2o_ppm * calcAirDensity(press, temp) * 1.0e-06_DP
+    ! Eq.18 (see Vaisala paper). We don't use ppm here, because doing
+    ! so requires a multiplication by 1e6 then division by the same
+    ! in the final line. This incurs possible rounding. Instead, we
+    ! use 'parts per unit'
+    h2o_ppu = wvp / (press - wvp)
```
spco: @rs028 - I hope the reason for this change makes sense - having values of around 1e11, if we multiply by 1e6 and then by 1e-6, we lose precision as we can only hold about 16dp in `_DP` reals. I see this in practise on the "convertRHtoH2O: 50% relative humidity, 0C, pressure = A" unit test below, but making this change ensures we don't see such an effect. As `h2o_ppm` is not passed out of the function, it doesn't matter to the user.

PR #310, opened 2018-08-01
Use caching of gfortran on Travis + MacOS.

spco:
This should save about 2 minutes on the Mac Travis build, which would. be useful!

As SIP blocks the export of DYLD_LIBRARY_PATH to scripts, we use a new FORT_LIB variable in the Makefile to pass on the locations to use. Also make sure PATH is exported, not just set on .travis.yml, and run make behaviourtests after make unittests, rather than make test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,15 +24,16 @@
 
   before_install:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export DYLD_LIBRARY_PATH=$PWD/cvode/lib:$PWD/openlibm-0.4.1 ; fi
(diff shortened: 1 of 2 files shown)


spco:
The merge to master fails, because the cache has not been built. I'm going to have to add one PR that creates the cache, and another that adds the use of the cache. Sigh...

rs028:
@spco cool. I know it is a minor point, but can we call the `behaviourtests` in another way? I always get the spelling wrong :(

spco:
I agree it's a bit of a mouthful. How about we have `make tests` and `make unittests`, and both are called if you `make alltests`?

rs028:
much better :)

PR #311, opened 2018-08-01
Revert "Use caching of gfortran on Travis + MacOS."

spco:
Reverts AtChem/AtChem2#310


spco:
Reverting is needed because we need one commit on master to generate the cache, then one to use it. Using one PR for both those stages only results in the latter commit being tested, but this fails before te end of the `script` stage of the travis file. So we need to have one PR to generate the cache, then another to use it.

PR #327, opened 2018-08-09
add header and reaction numbers to instantaneous rates files.

spco:


rs028:
@spco not that it matters, but why does it put so much space before the numbers?

spco:
Because I forgot to strip out the leading whitespace! I'll push out a fix for that.

rs028:
ah okay, I  thought I was misunderstanding the code :)
if you are making changes to that I have another suggestion. For clarity I think that if the header is  `reaction number : reaction rate` than we want the `:` separating the numbers as well. Alternatively, blank space is fine, but then I woul dremove the `:` from the header.

spco:
I ummed and ahed about this for a while. If I use a whitespace in the header, then it just says

reaction number           reaction rate

which is not very clear. I don't particularly like 

reactionNumber reactionRate

but at least that would be clearer, and more consistent with other output files.

I'd caution against sticking a : in each line - that is just going to make it more difficult to parse the data into any plotting program. I think this leads into a larger discussion about what format is best to output into - better to output in a useful form than ask users to write complicated code just to wrangle information out of the data format.

I'm happy to move to 
reactionNumber reactionRate
for now if you like, as that's consistent. The larger discussion is definitely for the future.

rs028:
Okay I see your point. Let's do ` reactionNumber reactionRate` for now. Eventually, I want to think about how to have better output. I feel the space delimited text files may be a bit limiting as a format.

spco:
Handled by #328

Issue #305, opened 2018-07-13
Fix  calculation of currentFracHour


rs028:
Small error in the calculation of `currentFracHour` affects the calculation of the solar angles.

spco:
This is documented at https://github.com/AtChem/AtChem2/pull/266#pullrequestreview-107428474.

spco:
`currentDayOfYear` is calculated only at the start of each _outer_ timestep.

The callgraph for `calcZenith()` is 
```
mechanism_rates(t,y,p) -> getEnvVarsAtT(t) -> calcZenith(t, dec)
```
which means that the `t` in `calcZenith()` is the time at the _inner_ timestep. I'm not sure whether this causes the misalignment seen in the link above. By this I mean that it's possible that `t` has moved on into the next day within a single call to `CVODE()`, but this won't update `currentDayOfYear` as this is only updated once per _outer_ timestep.

From what I can see, I don't think this should make any difference to the calculation of quantities such as `lha` or `cosx`.

Have you seen an example where this definitely affects these quantities, or is it that the misalignment makes you suspect that the later quantities may be affected?

spco:
Can I clarify that the time we use is always GMT, wherever the actual lat/long posiiton of the simulation is? This is what is being used in the calculation of LHA, but it too me a while to figure out. I can't see this documented anywhere, which would definitely be needed by users.

rs028:
Yes it is always GMT. I think it is easier than dealing with local time (not to mention daylight saving) but I haven't really tested it for a range of lat/long.

spco:
Closed by #331

PR #331, opened 2018-08-10
Solar cleanup

spco:
Address issues with solar calculations. In particular, split `eqtime` calculation into a new function, simplify the `currentFracHour`, and correct the calculation of theta for a zero-based day-of-the-year.


spco:
Now, `EQT` matches the reference calculations exactly, apart from a similar difference at midnight as seen with `DEC`. Both are now fine - I don't think the choice of `DEC` at exactly midnight is anything to worry about.

@rs028: My `LHA` values are still a little out, but this is a position-dependent value. What longitude are the reference values calculated for? I'm calculating for `lat=51.51, lon=0.13`, which could easily account for any discrepancy in my numbers, if you're calculating for 0 longitude. I can re-run with whatever longitude.

rs028:
I agree about `EQT` and `DEC`. Does it also deals correctly with the Dec 31-Jan 1 transition?

I calculated the values for lat=51.51, long=0.13 and dates "01-03-14 00:00:00" to "01-03-16 00:00:00".
How much out is it?

spco:
Yes, that transition works fine (apart form the midnight discrepancy as ususal).

Ok, looks like there's still something going on with `LHA`. We're a few percent out, but not a steady amount that I can see. There's something still to figure out about the differences.

spco:
Based on the spreadsheets at https://www.esrl.noaa.gov/gmd/grad/solcalc/calcdetails.html, our calculations here are at least as accurate those in calculated.csv from your email. And I'm happy to say that I was slightly misreading my spreadsheet before - we get agreement to within 2% relative error everywhere apart from at the midday value - these have very small absolute values, so the relative error gets quite large. But the absolute error hovers around 0.007 which is very small. So I think this may be sufficient - we have hit the point where the differences in our choices of approximation are the biggest factor, rather than any calculation issues.

rs028:
Excellent, great job!

spco:
Fixes #305

PR #332, opened 2018-08-14
Rename config and output files

rs028:
The new names are more consistent and it's more obvious to relate the configuration to the output.


rs028:
@spco, besides the codecov error message, there are bunch of warnings from FRUIT in the Travis log.
BTW, codecov seems to be running only on the Linux version is this expected?

src/inputFunctions.f90
```diff
@@ -269,14 +269,14 @@ end subroutine readAndSetInitialConcentrations
...
-  subroutine setConcentrations( concSpeciesNames, inputConcentrations, refSpeciesNames, outputConcentrations )
+  subroutine setConcentrations( concSpeciesNames, inputConcentrations, refSpeciesNames, outputSpecies )
     use types_mod
     use storage_mod, only : maxSpecLength
     implicit none
 
     character(len=maxSpecLength), intent(in) :: concSpeciesNames(:), refSpeciesNames(:)
     real(kind=DP), intent(in) :: inputConcentrations(:)
-    real(kind=DP), intent(out) :: outputConcentrations(:)
+    real(kind=DP), intent(out) :: outputSpecies(:)
```
spco: I'm not sure that this is a clearer name? The array contains the concentrations of all chemical species. `outputSpecies` suggests it contains a list of species.

Or is this just a find-and-replace error from the renaming of `outputConcentration.config`?

spco:
Yes, I need to submit a PR to fix those warrnings. It does indeed only cover Linux - there shouldn't be any difference between the two, and it speeds things up, so I think it's fine.

src/inputFunctions.f90
rs028: find and replace error :)

rs028:
@spco this has been prompted by a number of user feedbacks I have received. I am happy with it now. if you have the time to quickly check that it's all good I will merge it.

spco:
Looks good to go. Would it be better to squash the commits, as several of these individual commits break on their own?

rs028:
You mean the `squash and merge` button on this page, right?

spco:
yep! Unless you want to faff around with doing it locally, but there's little point.

rs028:
No, that is what I do usually anyways (or is it wrong and would be better to do `rebase and merge`? I am not entirely clear of the difference, tbh)

spco:
Rebase and merge adds all the commits as they are to the top of `master`. Squash and merge squashes all the commits into one commit, then adds that single commit to the top of `master`. Generally, if you've got a PR containing a couple of commits that all passed and are meaningful, then rebase and merge is preferable. But if you've got lots of messy commits then it's best to squash and merge.

P.S. in the messy case there is an argument that you should clean up the local branch so that it's made of nice commits, then force-push to the PR and then use rebase and merge. This means your whole PR isn't squashed into one commit, but the individual commits are nice going into `master` too. I often do that, but I won't force you to!

PR #334, opened 2018-08-16
Solar tests

spco:
This contains unit tests for all the functions in `solarFunctions.f90` bar `calcZenith()`. I think that function wants refactoring, so I've left that as a TODO. `calcDec()` now has a helper function `decFromTheta()` which takes out some of the guts, and makes unit testing simpler.


spco:
Note to self: Requires squashing when merging

PR #336, opened 2018-08-23
Add config tests

spco:
Currently just missing getConcForSpeciesOfInterest(). Includes fixing the result of findReactionsWithProductOrReactant() (this was out by one in the lengths). Fixes a style issue in outputFunctions.f90.


spco:
Second commit fills the hole in the unit tests - all configFunctions.f90 functions now have unit tests.

PR #339, opened 2018-08-24
Unit tests for constraintFunctions.f90

spco:
Add unit test for calcPhotolysisRaw


rs028:
@spco do you want to merge this as well, or there is still work to do here?

spco:
Yeah, this is a bit of a stub currently, and requires a fair bit more work, so I'd say leave it for later.

rs028:
okay, no problem. We are good to go then.

PR #340, opened 2018-08-29
Add new tests

spco:
Add static test to ensure a simple symmetric two-species setup is stable over a long time period. This is the first of the new-look tests.
@rs028: feedback welcomed - I've run it for a reasonably long time period to check we don't lose any precision.


spco:
https://github.com/AtChem/AtChem2/pull/340/commits/c7f1bca1f028e4a5b27fd38b63d62b36db2f6903 contains the system corresponding to a single first-order ODE pair:
```
A->B with rate k, A(0) = 100, B(0) = 0
```
where `k = 0.01*(281.95/300)` with solution
```
A = 100 exp(-kt)
B = 100 (1-exp(-kt))
```
The results on my Mac are shown here (black is calculated, red is exact solution. O = A, O3 = B, but here I've plotted (O3 - 100) so that we see the details of the later part).
oo3
It looks generally very good, but it dies at about 1216s. The points off the graph are actually negative concentrations. From my explorations, I think this is because `FCVFUN()` is using the value of `y` (the concentrations) from the beginning of the last outer iteration, (which can be seen by outputting instantaneousRates for every second - it only updates every few timesteps, even though the underlying concentrations have changed). Ideally, it would instead be using the very latest calculated concentrations.

This may or may not be possible - I'm not certain that it is. I'd need to look into it further. And it's not unreasonable to expect the calculation to struggle at 1e-6 - squaring that gets you down to machine precision. Either way, this test demonstrates that there's nothing fundemantally wrong with our calculations at a reasonable scale for this example - it matches the calculated solution.

rs028:
If I understand what you are saying this means that the last few timesteps of a run are not reliable? What happens for a very short run then? For example I have used the model for 100 seconds simulations: it's a special case but may be useful to understand this behaviour better.

travis/tests/static/model/configuration/static.fac
```diff
@@ -0,0 +1,187 @@
...
+*;
+* Variable definitions.  All species are listed here.;
+*;
+VARIABLE
+ O O3 ;
+****************************************************** ;
+*;
+* Generic Rate Coefficients ;
+*;
+KRO2NO = 2.7D-12*EXP(360/TEMP) ;
```
rs028: A minor point. It may be better to use the minimal mechanism (https://github.com/AtChem/AtChem2/wiki/2.1-Chemical-Mechanism#minimal-mechanism-file) for these tests, if only for clarity. and substitute the standard MCM header with a comment of that the test does.

travis/tests/static/model/configuration/static.fac
```diff
@@ -0,0 +1,187 @@
...
+*          If any of these are peroxy radicals the RO2 sum will be wrong!!! ;
+****************************************************** ;
+*  ;
+RO2 = HOCH2CH2O2 + BUT2OLO2 + CH3O2 + CH3CO3 + HCOCH2O2 + MEKAO2 +
+      MEKBO2 + MEKCO2 + C2H5O2 + HOCH2CO3 + BIACETO2 + HO1CO3C4O2 +
+      CO2C3CO3 + CH3COCH2O2 + C2H5CO3 + HCOCO3 + HOC2H4CO3 ;
+*;
+* Reaction definitions. ;
+*;
+% (TEMP/300) : O = O3 ;
```
rs028: Just as a note you don't need to have temperature dependence (unless that is part of what you want to test, of course). `k` can be simply `0.9` in this example.

spco:
It's not to do with how far through the simulation it is - it's because the concentrations are so small that the fact that we're not using the very latest value makes a difference. In an ideal world `FCVFUN()` would use the latest value of the concentration and it would be fine. But because it only updates that concentration every so often, when the concentration is small then the rate of change of concentration makes it overshoot to below zero.

travis/tests/static/model/configuration/static.fac
spco: I left this in as it is doing no harm.

travis/tests/static/model/configuration/static.fac
spco: Yes, that might be cleaner.

rs028:
I see, basically if it too close to zero it goes nuts :)
I think whether this matters or not depends on how much it overshoots. If the "danger zone" begins at 1e-2 (from your plot) and it overshoots by a similar quantity then probably it doesn't matter because +/- 1e-2 is, for all practical effects, basically zero. But if it overshoots to -1e+2 or more than it may have an effect on some other species (although probably minor). Is this correct?

spco:
It's not so much the absolute values that matter, but rather the relative values. In one experiment, a concentration of 1e-2 might be negligible, but in another it might be the largest concentration of the lot, so very important.
In actual fact I think the most important quantity is something like the ratio between the concentration of species A and the product of the reaction rate and timestep. So `conc(A) : r*t`. 
Essentially, we are making an error by using `r(t_1)` throughout the interval `[t_1, t_2]`, when we should be using `r(t)` at all times. If the concentration `c` is low, then if you use the reaction rate at timestep `t_1` and step forward to `t_2`, then if `r(t_1) > c/(t_2-t_1)` then by `t_2` the concentration will be negative.
This affects most such numerical schemes, and is tackled by reducing the timestep, but I don't think that's an option for now.

And I wouldn't expect the scheme to get back on track - once accuracy is lost it can't be regained. Rather, the game to play is to make sure that you lose as little accuracy for as long as you can.

rs028:
Okay, I understand. Isn't this why we adopted `openlibm` as well? Is it all the same issue?

spco:
Further experimentation with the timstep lengths suggest it's not just the timestep length that is causing the problems at about 1200 seconds - it's a feature even with 0.01s steps. Instead, that suggests that a quantity around there (perhaps the product kAB) is hitting round-off problems.

rs028:
I understand. So how do you suggest we proceed? figure this out first or do more testing first?

spco:
I'm doing a little more testing to weedle out issues. If I have the system
`A + A -> B + B`
with rate `k`, can I confirm whether the reaction rate is `kA^2`? I'm doing some testing, but when I compare to the exact solution it's the exact solution to the system with reaction rate `2kA^2`. I want to make sure I have the right end of the stick before I spend more time working out what's wrong.

rs028:
see here https://opentextbc.ca/introductorychemistry/chapter/concentration-time-relationships-integrated-rate-laws-2/

spco:
After a little offline discussion, everything above is actually ok. The `secondorder` scenario tests the `A+A->B+B` setup, and works perfectly, as demonstrated:
secondorder

spco:
@rs028 - do you agree that this is ready to merge? It contains the 3 tests, with the .fac files squashed to a minimum. These will in time replace the existing tests, but for now they just augment them.

spco:
And if there are any other exact-solution situations you can think of (I've run out of ideas for now!) then do let me know.

travis/tests/exponential/model/configuration/exponential.fac
```diff
@@ -0,0 +1,26 @@
+********************************************************************* ;
```
rs028: should this maybe be called `firstorder.fac`, for consistency with the other test?

rs028:
I think for now this looks okay. Eventually we will need a second order test of the type `A+B = C` but the integrated rate law is a bit more complicated, so if you want to merge this now it's okay for me.

travis/tests/secondorder/model/configuration/secondorder.fac
```diff
@@ -0,0 +1,27 @@
...
+*;
+* WARNING: The following species do not have SMILES strings in the database. ;
+*          If any of these are peroxy radicals the RO2 sum will be wrong!!! ;
+****************************************************** ;
+*  ;
+RO2 = ;
+*;
+* Reaction definitions. ;
+*;
+% 0.0001*(TEMP/300) : O + O = O3 + O3 ;
```
rs028: Just to be clear. This case basically also checks that `A+A=B` and `A+A=B+C` are correct, right?

spco: While it doesn't explicitly do so, this case is probably "more complex" than either of those, so it provides some reassurance that they would behave as expected too.

rs028: That's what I thought. Very good. :)

Issue #259, opened 2017-11-21
Warnings related to mechanism file


rs028:
1. The MCM usually uses `D` to indicate the exponential but sometimes uses `E`. This results in a number of compilation warnings about single precision constants.

2. Sometimes the MCM extractor breaks long lines like this:

```
% 8.8D-12*EXP(-1320/TEMP) + 1.7D-14*EXP(423/TEMP) : CH3COCH3 + OH =
 CH3COCH2O2 ;% J<21> : CH3COCH3 = CH3CO3 + CH3O2 ;
```

These are problems with the web extractor and/or the MCM database, but need to make sure they are not problem for the python scripts and the compiler and they do not affect the model results.

rs028:
More info on the second point. The files downloaded from the MCM website seem to use a mixture of Line Feed and Carriage Return. This is apparent when using a text editor that shows new line characters.

For example emacs shows:
```
% 2*(KCH3O2*2.4D-14*EXP(1620/TEMP))@0.5*RO2*0.2 : C6H5CH2O2 = BENZAL^M
;^M% 2*(KCH3O2*2.4D-14*EXP(1620/TEMP))@0.5*RO2*0.6 : C6H5CH2O2 = C6H5CH2O ;
```

Where `^M` are the Carriage Return characters. Parsing the file as is with the python scripts gives the correct result. In `mechanism-rate-coefficients.f90`:
```
p(2) = 2*(KCH3O2*2.4D-14*EXP(1620/TEMP))**0.5*RO2*0.2   !% 2*(KCH3O2*2.4D-14*EXP(1620/TEMP))@0.5*RO2*0.2 : C6H5CH2O2 = BENZAL ;
p(3) = 2*(KCH3O2*2.4D-14*EXP(1620/TEMP))**0.5*RO2*0.6   !% 2*(KCH3O2*2.4D-14*EXP(1620/TEMP))@0.5*RO2*0.6 : C6H5CH2O2 = C6H5CH2O ;
```

However if the ^M are removed (for example if the mechanism file is modified and the text editor applies consistent new line characters on saving the file) the python script fails:
```
[...]

PR #343, opened 2018-09-13
Remove perl search for single-precision constants.

spco:
This is no longer relevant as constants use the types_mod-defined types - the search also picks up certain print format specifiers.


rs028:
Is this solving #259?

spco:
No, it just removes the warnings at the end of the `make`.

PR #346, opened 2018-09-17
Handle RO2 sum in normal source code (not generated)

spco:
This PR is a precursor to the function parser. This part doesn't need to be generated on the fly, and would be better handled by being treated by the python converter. This is done in this PR (when complete) to clear the way for the function parser.


spco:
@rs028 This PR is finished now - it removes the `ro2-rates.f90` files, and instead calls reads the ro2 speceis numbers from file at runtime. This also means `gen/mechanism-rates-doefficients.f90` doesn't include the call to `ro2sum()`, which was a legacy issue and can perfectly well live in the normal source files.

Let me know if you spot any issues with this.

rs028:
Looks good to me!

PR #347, opened 2018-09-18
Replace generic complex variables with vector q

spco:
This leaves us in a usable state, but much closer to using the function parser. `tools/mech_converter.py` has been extensively reworked to parse the .fac file much more finely. Now, each of the generic rate coefficients and complex reactions is held as an element within a vector `q`, rather than as a separate variable. The aim is that then we can just teach the function parser to treat `q` (and possibly `J()` too) as a special case.

This could probably still do with a little better documentation of the python code before squashing.


rs028:
@spco I am wondering if it would be good to have an official version **before** completing the switch to the function parser. The main reason is that, if we think that it might slow down considerably the model, it would be good to have a stable fallback option. Secondly, it is a major change and I think it makes sense to say "this is where we got to with the original code" and get a paper out with the original authors at that point. Third I'd like to clean and archive the documentation for future reference. What do you think?

spco:
That might well be a good idea, as it is quite a change, and for all the reasons you gave. I was thinking about the doecumentation only this morning - the wiki is not great for versioning (although it is actually a git repo in its own right I believe). There must hopefully be a decent mechanism for versioning that in an easy-to-access way.

Issue #319, opened 2018-08-06
Correct the lossRates and productionRates output


spco:
`solverFunctions:66-74` calculate lossRates and productionRates wrongly. Proposed solution:

Switch to use only one config file to replace `outputLossRates.config` and `outputProductionRates.config`. Call `outputRates.config`. Then `lossRates` and `productionRates` will refer to the same species.
Fix the equations.
Fix headers in `mechanism.prod`?
Check stoichimetric coefficient handling.

spco:
Output rate should be reaction rate divided by concentration of species of interest.

E.g.

`A+B--(k)-->C+D` has reaction rate `kAB`, so if `A` is species of interest, then `kB` is required.

Need to handle the cases `A+A -> B`, or `A->A+B`, `A->B+B` (and combinations).

spco:
I don't like the way `reacDetailedRatesSpecies` and `prodDetailedRatesSpecies` contain the number of the species in the first column and then the numbers of the reactions in the following columns. This should be split up. The current name is also confusing, and should be applied only to the first column when split up.

spco:
@rs028: a question on the desired outputs of lossRates and productionRates. Suppose I have the reaction `A+B ->C+D` with rate `k`. Then the reaction rate is `kAB`, where here and below I meanc 'the concentration of `A`' when i say `A`. I understand that if `A` is a species of interest then in `lossRates.output` there should be a line associated to it in this reaction, with value equal to `kB`. Similarly if `B` is of interest then we get `kAB`.

What happens if `C` is a species of interest? Should I expect a value in `productionRates.output` that is `kAB`? Or perhaps `kAB/C`? I just want to clarify, as that affects how loss rates are calculated versus production rates.

I'm ignoring the fact we want to later merge both these output files into one for now.

rs028:
Given the reaction `A+B -> C+D`:

- species of interest is A
production of A is : NULL
loss of A is : kB

- species of interest is B
production of B is : NULL
loss of B is : kA

- species of interest is C
production of C is :  kAB
loss of C is : NULL

- species of interest is D
production of D is :  kAB
loss of D is : NULL

where I am using NULL to say that, for that reaction, species X is not produced (or not lost). Does it make sense?

rs028:
Yes you are right. Keep in mind that those "reactions" are usually multiple reactions condensed into one for simplicity. What I was trying to do is to get the relative (to the species of interest) reaction rate, which can be  useful for some type of analysis.

[...]

PR #351, opened 2018-10-26
Change the outputs to loss/productionRates.output, as the previous values were meaningless.

spco:


spco:
This introduces the derived type `reaction_frequency_pair` so that we can use an array of these types to track the pairing of (reaction number, number of occurences). The large number of changed lines is because in the outputs (1) any reactions that had 2+ occurences of a species of interest used to print 2 lines - they now print just once with twice the rate; and (2) previously `lossRates` and `productionRates` often contained garbage, whereas we now print the values as given by https://github.com/AtChem/AtChem2/issues/319#issuecomment-416643702

rs028:
Brilliant! I need to use this output for a model so I will check the output.

I assume you also double checked the python script that processes the mechanism `fac` file in order to sort this out. Is #259 still an issue?

spco:
Hopefully this is all working now - let me know if you see anything wrong. I've tried to manually verify the changes to the output, but there's only so much I can do!
#259 remains an issue - I have worked on a fix this afternoon, which isn't quite ready. Once that is done then we're nearly there for v1.1!

PR #354, opened 2018-10-29
Consistent env var names

spco:
Always use 'press' and 'blheight' for the names of the environment variables. This matches currently used input config files.


spco:
Fixes 6. on #270. @rs028 let me know when you're happy with this choice.

rs028:
yes, that looks okay.

Issue #326, opened 2018-08-09
Un-useful tests


spco:
We already know that we can condense the existing tests down, or replace them with a better framework. This is just to note that the test `short`, `short_end_of_day`, and `short_no_pre` are all stationary models - the only non-zero concentrations they have are `CH3OH` and `NO2`, neither of which then appears in the accompaying .fac file. This isn't very useful.

rs028:
If I recall correctly all the `short*` tests are basically just testing the various possible configuration of the environment variables and reading in the contraints.

rs028:
Somewhat related question. Does each test need to have its own `mcm/` folder?

rs028:
An `INFO.md` file with the list and a brief description of each behaviour tests should be added (only for those we keep, not those we plan to delete and/or replace with unit tests).

rs028:
We also need a better way to log the results of the tests. At the moment one has to scroll up the terminal to see the results of the unit tests, for example, and this could be a pain if there are lots of tests. Maybe some of the output can be redirected to a log file (see also #320).

PR #357, opened 2018-11-01
Use main mcm/ directory for tests

rs028:
The `mcm/` folder does not need to be replicated in each test, I think (mentioned in #326).
@spco do you agree?


spco:
The only caveat I'd say is that what if we (perhaps in the future) want to implement a test that uses a different mcm version, perhaps one that isn't included in the standard directory? I mean, currently the version of the MCM we use is hard-coded (in multiple places!) in the code, so it's not exactly flexible in that regard. But that whole thing needs re-jigging anyway, as it's rather clunky to change the MCM version at this point. That can wait for v2. Up to you whether you think it's right to change to use the main MCM directory. Perhaps it is for now, if just for reducing the number of files floating around.

On a related note, we seem to have a `photolysis-rates_v3.2` but no matching `peroxy-radicals_v3.2`?

rs028:
I think that for the tests it should not matter which version of the MCM is being used. Each version of the MCM is different in terms of number of reactions and species but from a mathematical point of view it should not matter. The only place where I think it would make sense to check for different behaviour is the python mechanism parser, but one could write a test just for that, I suppose.
And as you say, the MCM version setting will need a different solution anyways at some point, so we can decide then if we need to run tests for multiple versions.

I know about the photolysis rates data. They are not on the website so I asked @pb866 if he can find them.

PR #361, opened 2018-11-08
mcm lists

rs028:
Correct RO2 list for mcm 3.1 and add photolysis parameters for MCM v3.2


spco:
Is this ready to go? 😃

rs028:
Did you see my note?

spco:
Not that I can remember - what note was that?

rs028:
In this PR. I asked you to check if the text in the new file `mcm/info,md` is correct :)

spco:
No idea where that note has gone (I can't see it anywhere?) but it all looks good. The last sentence is just missing a closing parenthesis.

rs028:
It appears on mine as "pending review", but okay it doesn't matter. If the text is correct, I will fix the parenthesis tomorrow (I am away now) and merge.

spco:
The Travis cache had expired, and the logic to rebuild it wasn't quite working. I've merged #362 to fix it - would you be able to rebase locally against the latest master so that you get #362? You'll then have to force-push to here.

rs028:
Not really. I am out of the country and for some reason I cannot login to the machine. I did the previous commits via the github web interface. I am not sure if I can rebase in this way. Are you able to do it? Otherwise it will have to wait until I am back.

rs028:
I managed to login and rebase but I think I messed up. I need a more stable connection!

spco:
Ok, it looks like my 'fix' didn't do the trick. Hopefully my second fix (#363) has now done the trick. Please rebase in the same manner as before (it looks fine to me, apart from the issue you identify about the readme.md).

rs028:
Hopefully this is now okay. Can you let me know if I left behind some bits and pieces before I merge it?

spco:
I can't see the original commits, only what's been force-pushed now, but it looks right to me - the correction to the parentheses, and the v3.1 and v3.2 files. I can't vouch for the `{INFO,CITATION,README}.md` files, as I don't remember what you added in detail. Good to see that it passes though 😄

Issue #369, opened 2019-01-07
Use dynamic loading of shared libraries instead of parser


spco:
PR #367 is showing fairly crippling performance issues, which are not easily addressed (the code is very difficult for the compiler to optimise). Instead, we should investigate the use of dynamic loading of shared libraries.

A rough outline:
- Convert `.fac` file to all the existing data files, but replace the generation of `mechanism-rate-coefficients.f90` with another file denoting a valid library.
- Compile this file into a shared library.
- Instead of using `include 'src/gen/mechanism-rate-coefficients.f90` in `mechanism_rates()`, call the function defined in the shared library (whose location will need to be an input argument).
- This code gets called as an external function, but using native hard-coded Fortran.

Benefits over the existing method: 
- Main AtChem executable needs compilation only once, with the data for a single model being held in a single location, including the original `.fac` file, the generated data files, and the shared library.
- Speed approaching that of a fully hard-coded solution - some IPO may not be quite as effective as in the previous case, but hopefully this shouldn't be too far off.

Benefits over function parser:
- Speed, hopefully by a huge factor.
- The source for the shared library will look very similar to that of the existing generated file - just with a little extra padding around it.
Possible cons:
- The shared library is another file to keep track of (though it should remain with the original data).
- Possible incompatibilities if the shared library is compiled with different versions to the main executable - but this should be solved by user education and use of a Makefile.

spco:
This is completed by #370

PR #370, opened 2019-01-08
Shared library

spco:
Addresses #369.

Convert the .fac file to a standalone Fortran module, which is dynamically linked as a .so to the main AtChem2 executable.


spco:
@rs028 I think this is ready for review. The only failures are from the code coverage, which I don't think I can do much about.

Work remaining: Update the README (and the wiki, once v1.1 docs are out) with the changes to how things run.

In particular, I'd value you thoughts on whether the new shared library should normally go into `model/configuration` (or `travis/tests/testname/model/configuration` for tests). Is this a sensible place? It matches where the `.fac` lives for the tests, but the build test uses `tools/mcm_example.fac`, so the `.so` ends up apart from the `.fac` (but with the other config files, including the other `mechanism.*` files).

rs028:
Good question. I tend to put the `.fac` file in `model/` rather than `model/configuration` so that the latter contains only `.config` and `.parameters` files. I have no problem with the shared library residing in `model/` or in `model/configuration/`. Or there could be an additional directory `model/mechanism`.

With regard to `mcm_example.fac` I am wondering if `tools` is the right place for it. Maybe it should go into `mcm/` and be renamed `mechanism_test.fac`. That being said, is it possibile to force the shared library to be wherever the mechanism file happens to be? It would be the easiest solution.

spco:
Back to the matter in hand - yes, we can force the location of the shared library to that of the mechanism file, and I think that's sane behaviour (in fact the only reason this doesn't do that currently is because that wouldn't work for `mcm_example.fac`'s current location).

I can see an argument for the default being `model/`, as you're right that it's not really config in a natural sense. Perhaps `model/mechanism` might even be preferable? That way, the `.fac`, `mechanism.f90`, `mechanism.o` and `mechanism.so` are all bundled neatly together.

rs028:
This all sounds very sensible. Having a lot of arguments is a bit of a pain, but also can cause to errors. With or without a smarter command line interface, it may be a good idea to simplify and nudge towards good behaviour.

Possibily the easiest solution is to decide that the directory `$MODEL` must have 4 sub-directories with specific names: `configuration/`, `constraints/`, `output/` and maybe `mechanism/`. In this way the user only needs to specify 1 argument. But maybe that would be too rigid.

This is probably a different PR? Or needs to be resolved now to finalize this one?

spco:
I'd suggest if you're happy for now, we merge this PR and open a new one. I might even have a look at improving the interface. Are you happy if this PR goes ahead?

rs028:
Yes, okay.

spco:
Hi @rs028 - I'm back from leave so looking to wrap this up. Are you happy if I rebase this and merge into 1.2-dev (which may want to be 2.0-dev in the end)? I can also open the other PR for the restructuring as discussed above.

rs028:
Welcome back :) Yes, I am happy to do so.

spco:
@rs028 Before I commit this, I want to update (a) the README.md to reflect the change in the build process. I will also want to update the documentation. Should I update the contents of `doc/` in this PR, and/or the wiki pages afterwards? Or are we using the wiki as the canonical live documentation, and `doc/` only contains the snapshot of the wiki at the latest release?

rs028:
Good question. My plan was to convert the files in `doc` to latex and then use those as live documentation. A pdf file can be attached to each new release.

This would make the wiki redundant, as it would be a bit cumbersome to keep it in sync with the latex files, but I am not sure what the best approach is.

Right now I have changed the `md` files only when I closed PR #375.

spco:
Ok, yes. I guess we should deprecate the wiki (either remove it, or add bold headers explaining that it's deprecated) in order to not confuse users.

https://github.com/palantir/godel/wiki/GitHub-wiki seems to be a useful tool (it updates the wiki to reflect the contents of `doc`), but remembering to run it frequently is probably a pain.

Perhaps the wiki could also be updated only at the time of release, like the LaTeX? Would that be useful to users?

rs028:
I am leaning towards removing it once the transition to latex is complete. Maybe keep one or two wiki pages with practical examples of usage that won't fit in the manual. What do you think?

spco:
Perhaps that's a sensible compromise - let's go with that. Point users to the main docs, but also supply some useful extras that may not be ideally suited within the main docs.

PR #373, opened 2019-01-09
Add documentation directory

rs028:
The wiki will be moved here. This should be merged before #370 to preserve complete documentation of v1.1.

@spco is it worth having a mini-release before #370?


spco:
Having a mini-release of v1.1.1 might make sense in order to have an easily-identifiable point in time for the documentation, and it will have to change a fair few bits to then match the behaviour from #370.

rs028:
I think so too. 

I am happy with the status of the wiki now. If you have 15 minutes can you please quickly review it, especially sections 2.5, 2.6 and 2.7? Just to make sure there are no errors/typos, etc... 

When you are done, I can move the `.md` files into `doc/`.

spco:
I've reviewed it and there's nothing that stands out (I corrected one typo in 2.7). It looks good to go from my view. Thanks a lot for all the effort you've put into that documentation - it's not a small piece of work!

rs028:
Yes, but I see the markdown format as a provisional thing. I think it would be better to convert to latex, so we can generate a pdf that can be attached to each release.
For the moment the wiki stays. The only question is whether we keep updating it or we just update the files in `doc`.

Issue #320, opened 2018-08-06
output of reaction rates


rs028:
Improve the format of the reaction rates output files

- `instantaneousRates`: change name to `reactionRates`, add column with reaction number, add header

- `lossRates` and `productionRates`: the current "3D" format is awkward to use.  Consider dropping some columns, reformat as 2D matrix. Maybe different files for each species?

rs028:
Cool. For point (2) I am not really sure what is the best approach, but I would leave it for now. I think the testing and numerical stuff should have priority.

rs028:
Note: the headers of `model/output/lossRates.output` and `model/output/productionRates.output` miss one column

spco:
> Note: the headers of `model/output/lossRates.output` and `model/output/productionRates.output` miss one column

This is addressed by #366

spco:
`initialConditionsSetting.output` is removed by #379

PR #379, opened 2019-02-01
Remove the use of initialConditionsSetting.output

spco:
This file is now redundant - it was really just for debug when we were less confident in the behaviour. As we don't see any problems now, there's no reason to keep the file.


spco:
Addresses part of #320

Issue #276, opened 2018-06-07
COSX threshold


spco:
The current value of cosx_threshold (1e-30) is equivalent to | x - pi/2  | < 10^{-30}, i.e. the sun is within 10^{-30} radians of the horizon, so that is probably not a useful threshold. This needs experimenting with to make sure the threshold kicks in before the numerics take a hit, but not too soon.

rs028:
Is this related to point 1 in issue #258?

spco:
Yep. I think the current logic is the following in pseudo Fortran:

Read from `environmentVariables.config`: 
```
select case(JFAC)
  case NOTUSED, set JFAC=1
  case CONSTRAINED, set JFAC directly from `environmentConstraints/JFAC`
  case CONSTANT, set JFAC to that constant
  case CALC, read in the JFAC species, and use the constrained date and photolysis factors for that species to calculate JFAC at time t
end case
```
So, if JFAC is CALC, then the photolysis rates should be set to zero at night, as long as our threshold is actually working. In all other cases, the photolysis rates are unaffected directly by time of day, but obviously a constrained JFAC or constrained photolysis rate may assert a dependence on time.

rs028:
I am wondering if this could be a bug. In principle, you would like a consistent behaviour with the various settings. If during nighttime (that is when SZA is above a given threshold) the photolysis rates go to zero, they should go to zero always, not only when JFAC is calculated. A few points:

1. Should this depend on the settings of JFAC or on the settings of the photolysis rates? The two settings are related but they can have a number of permutations. My feeling is that this should depend on the photolysis rates, regardless of what JFAC does.

1. The point of setting photolysis rates to zero (or extremely low value) is to eliminate the possibility of numerical artifacts during the night, when we assume that, without the Sun, these reactions cannot happen. So:

   - If the photolysis rates are calculated, does the algorithm set them to zero?
   - If they are constrained (all or just some of them), should the model override the constraints during the night? (the rationale being that any measurement in that period could be simply noise from the instrument).
   - If the photolysis rates are constant then they should never go to zero (obviously).

1. Also, JFAC calculation currently does not work very well so it can't be relied upon to do other calculations.

1. Keep in mind that what happens at sunrise and sunset is always critical because that's when the chemistry changes fast and mathematically it becomes more difficult. Therefore the transition to zero must be dealt with carefully.

[...]

PR #381, opened 2019-02-06
More sensible sun angle threshold

spco:
Update cosx threshold from 1e-30 to 1e-2 - which corresponds to 0.01 radians or 0.57 degrees.

Addresses #276


spco:
@rs028 are you happy for me to merge this? This is enough to stop me seeing the IEEE_UNDERFLOW_FLAG warnings that we see when running the basic model with the `tools/mcm_example.fac` mechanism. It may or may not also help with the numerical instability we have seen in the past.

spco:
Hopefully this is at least a more sensible threshold than the previous one, but it is probably still worth us discussing what the _proper_ threshold should be, or whether a different mechanism rather than a cutof should be used. I won't close #276 yet as it contains some other bits still to be worked through.

rs028:
Agreed. The proper threshold should be discussed in the context of issue #276 and requires careful handling.

PR #385, opened 2019-02-24
Convert doc to LaTex

rs028:
Convert the documentation from markdown to latex files.


rs028:
Needs some tidying up and cleaning of the latex code. The resulting `pdf` is not pretty but contains all the info in the wiki.

**Note:** is it good practice to have a binary (pdf) file in a git repo?

spco:
Looks good - yes, it could do with tidying up e.g. fixing the `[{[{` all over the place, replacing them with internal links.

While purists might worry, I've seen many people put a PDF into a git repo, and I think it's a good idea. We should probably make a decision as to whether the PDF should always be updated to reflect the .tex files (I'd vote for this).

Could you share here the compilation commands you're using for LaTeX, so it's easy to reproduce next time.

rs028:
Yes, internal links, tables, references are a mess right now. Pandoc works fine but the conversion is not perfect :)
I agree the pdf file should always be updated. The compilation command is simply:
`pdflatex AtChem2-Manual.tex`

Will add that to the manual/wiki at some point. I think I will merge this, because I don't have time to do further cleaning right now and I want to focus on the paper.

spco:
Ok no problem - if I get a spare moment I will do some tidying up of the links etc after you've merged this.

rs028:
I would put this as a lower priority tbh. It is something I can do next time I am bored on a plane :)

PR #388, opened 2019-03-04
Change all python to be version-independent

### LINKED ISSUE #387: Python version 3
All Python scripts are written for version 2.7
- upgrade to version 3?
- make scripts version indipendent?

spco:
Closes #387


spco:
@rs028 This is ready to go, with only 1 thing to be decided - is it right that I've upgraded the actual calls use to `python3`? It's now writeen version-independently, so it will run with either 2.7 or 3.x. Is it wrong to expect users to have access to a python3? I also haven't looked at whether this affects the documentation.

rs028:
Not sure I understand the question :)
 Has it been converted to python3 or is it version indipendent?

spco:
Apologies, that wasn't very clearly put.

Each of the python scripts will now run perfectly happily under either 2.7 or 3.x. But in the `.sh` files, we call them with `python`. I've changed that here to call them with `python3` instead.

Upon reflection, I think I should change back to using `python` - that way, it will work whether or not users have 3.x installed (or 2.7).

rs028:
Okay, yes that makes sense :)

PR #393, opened 2019-04-19
Organize documentation

rs028:
Clean up the manual latex files.

Also adds the poster presented at the ACM conference (Dec 2018) to the `doc/` directory


rs028:
This okay now. There is still some work to do but it is mostly cosmetic.

PR #395, opened 2019-06-23
New script for version markup

rs028:
This script facilitates the update of the version number in all files - before creating a new release and before starting a new development cycle.

Fixes leftover issue from #10 and #293.

### DIFF
--- a/tools/version.sh
+++ b/tools/version.sh
@@ -15,7 +15,9 @@
 #
 # N.B.: the script MUST be run from the base directory of AtChem2.
 
-VERS_OLD="AtChem2 v1.2-dev"
-VERS_NEW="AtChem2 v1.2"
+VERS_OLD="v1.2-dev"
+VERS_NEW="v1.2"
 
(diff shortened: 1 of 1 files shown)


rs028:
@scpo what do you think of this?
Also can you send me an email with your new your new email ? I tried to send you the paper draft but it bounced.

spco:
I haven't tested this, but it looks good to me otherwise. A useful tool!

spco:
WARNING: When I actually run this on a Mac, issues with the differences in `sed` syntax corrupt my entire repository!

rs028:
Can you be more specific? I am not familiar with Mac/Linux differences in `sed`

spco:
Could we use perl instead of `sed`? `sed` on Mac doesn't allow inplace with the same syntax.

```
find ./ -type f ! -name "version.sh" -print | xargs perl -pi -e "s/$VERS_OLD/$VERS_NEW/g"
```

Does that work on your system?

rs028:
It is a bit slower but works. Thanks

Issue #389, opened 2019-03-08
organize tools/ dir


rs028:
Some more or less random ideas for the stuff in `tools/`:

- Does it make sense to move the build and compilation scripts out of `tools/` into their own directory in the main directory?

- ~Maybe move `mcm_example.fac` from `tools/` to `mcm/` (and rename `mechanism_test.fac`, seems more sensible choice)?~ [done by #396]

- ~Rename `Makefile` in `tools/` to something like `Makefile_skel` or `Makefile.skel` to make it obvious that it is not the real Makefile but a skeleton that needs to be copied to main directory and modified?~ [done by #396]

rs028:
All done by PR #396.

PR #396, opened 2019-07-17
Cleanup before the paper

rs028:
This PR contains a number of minor commits to clean up the repository, the manual and the wiki in advance of the publication of the paper.


model/configuration/environmentVariables.config
```diff
@@ -1,7 +1,7 @@
 1 TEMP			298.15
 2 PRESS			1013.25
 3 RH			NOTUSED
-5 H2O			3.91e+17
+4 H2O			3.91e+17
```
rs028: @spco Is this okay? Could it have caused issues in previous model runs?

spco: Good spot! But no, this number is unused - it's read at https://github.com/AtChem/AtChem2/blob/master/src/inputFunctions.f90#L827, and is an unused dummy variable, so we're safe :)

rs028:
@spco can #395 be merged?

spco:
@rs028, sorry, yes!

spco:
I haven't tested it myself, but if it doesn't break the Travis build then I think that's sufficient confirmation for me on this one.

rs028:
It doesn't locally, but I don't understand why Travis/Codecov has not started on this PR. Usually it is automatic.

spco:
It may be that the force-push meant the build was not triggered. Looks like the latest commit has triggered it!

rs028:
Yes :) btw, I figured out why there are two apps registered for Travis. The explanation is here https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/. It may be relevant for other projects you work on. It does not look like we have to do anything. The migration should be automatic (at least that is what I understand).

tools/build.sh
```diff
@@ -47,9 +47,15 @@ echo ''
 if [ -z $2 ]; then
   echo "make sharedlib"
   make sharedlib
+  echo "=> shared library directory: ./model/configuration/"
 else
   echo "make sharedlib" $2
   make sharedlib SHAREDLIBDIR=$2
```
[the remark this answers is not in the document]
spco: `if [ -z  $2]` returns True if `$2` is empty. So if no argument is supplied, `make` runs with the default `$SHAREDLIBDIR` within `Makefile`. If an argument is supplied here, `make` uses that instead of its own `$SHAREDLIBDIR`. So I don't see any reason to change the location, and doing so would mean you can't use `make` directly, and would be forced to always use `build.sh`.

tools/fruit_generator.rb
```diff
@@ -1,24 +1,36 @@
+# -----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
spco: Just seeing whether it will work on a re-run...

tools/build.sh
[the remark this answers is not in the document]
rs028: Okay, sounds sensible to me. So `model/configuration` is default for all `mechanism.*` files as well as for `mechanism.so`. It is just that the places were the default values are defined are different (`mech_converter.py` and `Makefile`). Right?

src/argparse.f90
```diff
@@ -274,13 +274,15 @@ subroutine get_and_set_directories_from_command_arguments()
...
+    write (*, '(2A)') ' Model directory is: ', trim( model_dir )
+    write (*, '(2A)') ' Output directory is: ', trim( output_dir )
+    write (*, '(2A)') ' Reaction Rates directory is: ', trim( reactionRates_dir )
+    write (*, '(2A)') ' Configuration directory is: ', trim( configuration_dir )
+    write (*, '(2A)') ' Constraints directory is: ', trim( constraints_dir )
+    write (*, '(2A)') ' Environment Constraints directory is: ', trim( env_constraints_dir )
+    write (*, '(2A)') ' Photolysis Constraints directory is: ', trim( photo_constraints_dir )
+    write (*, '(2A)') ' Species Constraints directory is: ', trim( spec_constraints_dir )
+    write (*, '(2A)') ' MCM directory is: ', trim( mcm_dir )
     write (*, '(2A)') ' Shared library is: ', trim( shared_library )
```
rs028: This is not pointing to a directory, but to a file, right?

it is `--shared-lib=model/configuration/mechanism.so`

not

`--shared-lib=model/configuration/`

tools/fruit_generator.rb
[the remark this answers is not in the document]
rs028: It does not look like

tools/build.sh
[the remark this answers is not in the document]
spco: Yes. The two are independent, but just happen to have the same default location/.

src/argparse.f90
spco: That's correct. As it's only ever a single file needed, not a collection, I think it makes more sense to point directly to it - this also allows you to put many `mechanism1.so`, `mechanism2.so` in the same directory if you want to.

tools/fruit_generator.rb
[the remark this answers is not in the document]
spco: Ah, I understand! In Makefile line 86, we replace the contents of line 4 in this file on the fly, in order to point at the correct `rake_base.rb` file. That needs to change to be line 17 now. Could you do that, rather than me? Just change the `sed -i "4s,.*` to `sed -i "17s,.*.

tools/fruit_generator.rb
```diff
@@ -1,24 +1,36 @@
...
+# Copyright (c) 2017 Sam Cox, Roberto Sommariva
+#
+# This file is part of the AtChem2 software package.
+#
+# This file is covered by the MIT license which can be found in the file
+# LICENSE.md at the top level of the AtChem2 distribution.
+#
+# -----------------------------------------------------------------------------
+
+# This script ...
```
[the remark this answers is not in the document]
spco: "This script creates a new FruitProcessor object, and generates files to control the unit tests."

Obviously if that needs a second line then the 17 above may need to become an 18 etc.

rs028:
Commit  cc26992 addresses the last open issue of #389. I think it is a bit more tidy this way. Would it be better to have `build/` in the main directory, instead?

spco:
On that point, I don't mind too much. It will be more tidying up for you to do! `build/` is often used as the location where intermediate files within compilation are put, but I'm not too fussed.

rs028:
We use `obj/` for that, right?

spco:
Yes, sounds about right.

tools/install/Makefile.skel
```diff
@@ -65,6 +65,7 @@ FORT_LIB     = ""
 endif
 endif
 
+## set compilation flags for each compiler
```
rs028: Something in these changes causes the unit tests to fail on Travis, but they work locally. Any clue?

tools/install/Makefile.skel
```diff
@@ -13,17 +13,27 @@
...
+
+## default location of the chemical mechanism shared library (mechanism.so)
+## use the second argument of the build script to change $SHAREDLIBDIR
+SHAREDLIBDIR = model/configuration
+
+## ================================================================== ##
+
 .SUFFIXES:
 .SUFFIXES: .f90 .o
 .PHONY: all test
```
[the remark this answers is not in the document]
spco: The `.PHONY` is standard in Makefiles, and necessary here - it tells make that it shouldn't expect to find any files called `all` or `test` being created by the rules with those names.

tools/install/Makefile.skel
```diff
@@ -16,16 +16,23 @@
 #
 # Makefile for AtChem2
 
-## paths to the dependencies
-## N.B.: use the full path, not the relative path
+# choose fortran compiler:
+# 1. "gnu" for gfortran
+# 2. "intel" for ifort
```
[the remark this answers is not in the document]
spco: We don't.

tools/install/Makefile.skel
```diff
@@ -45,10 +53,22 @@ else
...
 FORT_LIB     = ""
+FFLAGS       =  -fprofile-arcs -ftest-coverage -ffree-form -fimplicit-none -Wall -Wpedantic -fcheck=all -fPIC
+FSHAREDFLAGS =  -ffree-form -fimplicit-none -Wall -Wpedantic -Wno-unused-dummy-argument -fcheck=all -fPIC -shared
+LIBDIR       =  /usr/lib/:$(CVODELIB)
+endif
+ifeq ($(FORTC),"intel")
+FORT_COMP    = ifort
+FORT_LIB     = ""
+FFLAGS       = -free -warn
+FSHAREDFLAGS =
```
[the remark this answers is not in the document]
spco: The `ifort` behaviour hasn't ever been tested that I know of, but I built it into the Makefile at the time. I may be able to test it next week. The flags are likely to be different from the gfortran version, as they share very few identical flags.

tools/install/Makefile.skel
spco: L1307 at https://travis-ci.org/AtChem/AtChem2/jobs/561733054 indicates that it's not the Makefile, but rather than Fruit isn't being installed, because the download is failing. Looks like they've got an expired certificate. I would hope this will be fixed in the next few days :/

rs028: Ok, so it is their fault 😀 

What do you think of these changes? Do they make sense? I also wanted to add a bit more comments to the `Makefile` for future reference but I am not sure about all the sections.

tools/install/Makefile.skel
```diff
@@ -16,16 +16,23 @@
...
+# paths to the dependencies
+# N.B.: use the full path, not the relative path
 CVODELIB     = cvode/lib
 OPENLIBMDIR  = openlibm-0.4.1
 FRUITDIR     = fruit_3.4.3
 
-## default location of the chemical mechanism shared library (mechanism.so)
-## use the second argument of the build script to change $SHAREDLIBDIR
+# default location of the chemical mechanism shared library (mechanism.so)
+# use the second argument of the build script to change $SHAREDLIBDIR
```
spco: could you change this to "to override $SHAREDLIBDIR" as that's a bit more specific.

tools/install/Makefile.skel
```diff
@@ -16,16 +16,23 @@
 #
```
spco: Please move the #!/bin/sh to the very first line above the copyright, as that's the only place the kernel will ever look.

tools/install/Makefile.skel
spco: I've made review - just a couple of things really, the rest looks good and useful.

tools/install/Makefile.skel
rs028: Ok, that is also for all other bash scripts then.

spco: Ah yes, true!

spco:
It doesn't look like https://kent.dl.sourceforge.net/project/fortranxunit/fruit_3.4.3/fruit_3.4.3.zip has had its certificate updated as yet, so it may be worth moving to https://netix.dl.sourceforge.net/project/fortranxunit/fruit_3.4.3/fruit_3.4.3.zip. Addressing #399 should then make it easier to spot any future occurrences.

travis/run_tests.sh
```diff
@@ -12,9 +12,11 @@
 # -----------------------------------------------------------------------------
```
rs028: Some scripts have this `#set -e` command under bash. Should they all have it or not?

rs028:
@spco I tidied up and added comments to the Makefile. Can you tell me if it makes sense and correct/add the comments where necessary please?

travis/run_tests.sh
spco: `set -e` means it should exit if anything gives an error. I presume there was a reason why I put this in at the time, but I don't remember the specifics now. I don't think it's much of an issue in general whether the scripts have it or not - I'd personally leave it as it is.

tools/install/Makefile.skel
```diff
@@ -110,25 +114,30 @@ $(UNITTESTDIR)/fruit_basket_gen.f90 : $(unittest_code)
 	@cp travis/fruit_generator.rb $(UNITTESTDIR)
 	@cd $(UNITTESTDIR); sed -i "18s,.*,load \"$(FRUITDIR)/rake_base.rb\"," fruit_generator.rb; ruby fruit_generator.rb
 
+#
```
spco: build fruit_driver.exe from the individual unit tests

tools/install/Makefile.skel
```diff
@@ -76,29 +76,33 @@ FSHAREDFLAGS =
...
 OBJ = obj
 SRC = src
 UNITTESTDIR = travis/unit_tests
 
+# executable
 AOUT = atchem2
 
-all: $(AOUT)
-
+#
```
spco: UNITTEST_SRCS contains all source files except atchem.f90, as atchem.f90 is not unit-tested.

travis/run_tests.sh
rs028: it is only used intwo scripts: `travis/run_tests.sh` and `build/build_atchem2.sh`. But in the first script it is with a `#` and in the second without. Which one is correct?

spco: You're absolutely right. The `#` renders this one useless anyway, it's just a comment and can be removed. Let's keep the one in `build_atchem2.sh`.

Issue #355, opened 2018-10-29
Improvements to environmentVariables.config


rs028:
- [ ] Make sure that H2O and RH are mutually exclusive (if one is constrained or constant the other must be calculated).

- [ ] Make sure that the text in `environmentVariables.config` is unequivocal and intuitive (eg ROOFOPEN=ON is a confusing, ROOF=OFF is better).

- [ ] Simplify options in `environmentVariables.config` where possibile (eg, JFAC=1 and JFAC=NOTUSED are the same).

- [ ] What happens if there is a typo (eg, "CONTRAINED" instead of "CONSTRAINED") or an option that is not allowed (eg, "NOTUSED" for `TEMP`) in `environmentVariables.config`?

spco:
Looking at (5) about `DILUTE`, this is a subtle issue.

If `DILUTE`, `BLHEIGHT` or `RH` are set to `NOTUSED`, they end up having `envVarTypesNum = 4`. This is then handled on l316 of `constraintFunctions.f90` in `getEnvVarsAtT()`. Here, they are set to a 'nonsense value' of -1, but no error is thrown.

Ideally, we'd have an error triggered here if `DILUTE` is set to `NOTUSED`, but we're trying to evaluate it.

The difficulty comes that in the current implementation, we _always_ evaluate _all_ of the environment variables at each time of calling `getEnvVarsAtT()` - this function has no knowledge of whether our mechanism will actually make use of these values, so it just sets some nonsense. It's up to the user to ensure they never actually use that nonsense value - the idea of using -1 is that it should hopefully give weird enough results that the user takes a deeper look and spots the error.

I'm not quite sure how to easily handle this for now - we'd need to do something where either (1) the mechanism is checked to ensure it matches the `environmentVariables.config` and doesn't use a `NOTUSED` variable; or (2) somehow inform `getEnvVarsAtT()` about the contents of the mechanism. Neither is simple.

This also kind of raises the question of _why_ we'd ever want to put `NOTUSED` for `DILUTE` in the `environmentVariables.config` - if our mechanism file demands we use `DILUTE`, is the `NOTUSED` just to act as a warning to the user?

rs028:
So i think the weirdness in the data may be too subtle depending on how `DILUTE` is used but I also think we don't need to try to pre-check here because it is clearly the user who made a mistake  and we can't second guess what they wanted to do.

How about instead of a number we use a character, such as NA? So it the model tries a mathematical operation on it, it will crash.

[...]

PR #408, opened 2019-10-24
Dilute fix

bsn502:
New files: ./dilute.py
Modified files: ./build/build_atchem2.sh

The new code uses the "DILUTE" number in environmentVariables.config to amend the mechanism provided. Tests have been successful and the code is very useful, especially when using a larger mechanism file.

Current issues: the model needs to be re-compiled if changes are made to the dilution rate; the changes currently only work if model folder is called "model". 

An example using the ethane subset from the mcm and an initial concentration of ethane shows how its concentration changes under three different DILUTE rates.

dilute_test_v2


rs028:
This partially addresses the discussion on `DILUTE` in #355. It may eliminate the need to set this variable to `NOTUSED` and we can force the user to set its value to either zero or a dilution rate. If all species are diluted to zero this effectively means there is no dilution. However, this approach doubles the size of the chemical mechanism so the question is: can it slow down the execution? @spco any thoughts?

dilute.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+
+for x in spec:
+
+    x = x.split()
+
+    z = (x[1])
+
+    f.write(f"% {y} : {z} = LOSS ;\n")
```
spco: Would `= DILUTE` make it more obvious where this line has come from?

spco:
Actually, the more I think about this, the more I think this could be changed to avoid the double call to `mech_converter.py`, and make it cleaner overall. If the contents of `dilute.py` can be added directly to `mech_converter.py`, that makes everything a lot neater.

- Loop over `speciesList` to add the dilution reactions directly on about line 405 of `mech_converter.py`. 
- You would need to add another input argument to `mech_converter.py` to point to the `environmentVariables.config` file to get the value of `DILUTE`. Put it second (behind the argument for the `.fac` file) as the others are optional. 
- Edit `build_atchem2.sh` to fit that definition.

This removes the need to call `mech_converter.py` twice (which makes the flow easier to understand) and also removes the need to make a copy of the `.fac` file (and the resulting cleanup). Does that make sense?

build/build_atchem2.sh
```diff
@@ -1,60 +1,72 @@
...
-  make sharedlib SHAREDLIBDIR=$2
-  echo "=> shared library directory:" $2
-fi
-
-echo ''
-echo "make base executable"
-make
-
-exit 0
+#!/bin/sh
```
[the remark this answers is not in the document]
bsn502: I've made this change but there still appears to be a problem. There is a syntax error on the curly bracket of line:
f = open(f"./{mech_file}","a+")

I've come across this exact error when trying to use python 2.7... is Travis compatible with python 3?

rs028:
> (1) This seems like it will only work for a fixed value of DILUTE, but not for a constrained value. Is that an issue? I don't see a simple way to extend this mechanism to a constrained value.

At the moment `DILUTE` can be a number, `NOTUSED` or constrained. To be honest, I can't see a situation where `DILUTE` needs to be constrained, so I am not too worried about this now. That being said it doesn't hurt to have this possibility. The question is whether this approach (even if re-implemented as you suggest in https://github.com/AtChem/AtChem2/pull/408#issuecomment-546301638) can work or not when `DILUTE` is constrained.

> (2) Currently this only checks for `NOTUSED`. Could we also check for 0? If `DILUTE` is `0` (also handle `0.0` and `0.0e0` etc by converting the string to real), then we don't want to bother adding all these extra rows, in just the same way as `NOTUSED`. That will mean there's no overhead in the common can where no dilution is denoted by `0`. Thinking about it, we could go further and outlaw `NOTUSED` as a valid value - insist that the input is `0` instead. This would require modification to the Fortran parsing of the environment variables, but simplify the possibilities for users.

I agree. This is sensible. How hard is it to remove the `NOTUSED` option? Then `DILUTE` can only be one of these values:

1. `0`, in which case `dilute.py` is **not** called and nothing is added to the mechanism
1. a real in which case `dilute.py` is called and the dilution equations are added to the mechanism
1. constrained in which case `dilute.py` is called and the dilution equations are added to the mechanism but the dilution rate is read from a file in the environmental constraints directory (would it work? if not it's not essential, see above comment)

Does it make sense?

build/build_atchem2.sh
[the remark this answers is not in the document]
spco: Travis can use Python 3, but I think by default it's using 2.7 as that's what most systems will use if you just use `python` rather than `python3`.

We don't stipulate that users need python3 - it's better to maintain compatibility, hence for example the use of `from __future__ import print_function` in `mech_converter.py`. It may be useful to not use the `f" "` syntax, as that is python3-only, to maintain 2.7-compatibility. It would also match the syntax we use elsewhere in e.g. `mech_converter.py`.

Perhaps use the syntax (also using the `with` syntax which is more modern):
```
with f as open(mech_file,"a+")
    for x in spec:
        x = x.split()
        z = (x[1])
        f.write('% ' + str(y) + ' : ' + str(z) + ' = LOSS ;\n')
```

dilute.py
```diff
@@ -0,0 +1,68 @@
+import sys
```
rs028: Can you add some comments to help us understand the carious parts of the code?

Also, I think it would make more sense if this is inside the `build` directory.

spco:
> How hard is it to remove the NOTUSED option?

Good point. I don't think it would be too hard to remove the `NOTUSED` option - it's mostly just the handling at https://github.com/AtChem/AtChem2/blob/a74ec610bfa3ecb5d32fda1e484ded345989cb4d/src/inputFunctions.f90#L869 that would need changing.

But, that actually leads us on to thinking about what `DILUTE` now is. Can it be references like a species in other chemical reactions? I'm guessing not. In which case, it's no longer handled in the same way as the other environment variables (EVs), and should in some ways be removed from a lot of the EV functions. 

Currently, the whole process ends up with a value for `dilute` which can be used in the mechanism, but if it's just going to be used as a universal factor on all species as implemented in this PR, then we're no longer interested in re-evaluating that in the same way as the other EVs. So we might end up ripping a fair amount out (but just the lines that handle `dilute`).

To extend this to handle a constrained dilute, I think we can do something similar to what's done for the photolysis rates `J<1>` etc. Rather than printing 
```
0.1 ; O2 = LOSS
0.1 ; H20 = LOSS
0.1 ; CO2 = LOSS
```
etc, where `DILUTE = 0.1` in `environmentVariables.config`, instead we could print 
```
dilute ; O2 = LOSS
dilute ; H20 = LOSS
dilute ; CO2 = LOSS
```
and then evaluate `dilute` as we currently do at each timestep. Does that make sense?

spco:
Oh, and in fact, rather then the RHS saying `LOSS` or `DILUTE`, shouldn't it just be a blank, i.e.
```
dilute ; 02 = 
```

dilute.py
spco: Or see my comment about folding all of this into `mech_converter.py` :)

spco: But comments and removal of all those blank lines is good

spco:
Ok, please see https://github.com/AtChem/AtChem2/pull/412 for what I believe is a working version. It may be helpful to manually grab that commit's contents and use it to replace the contents here (that way we easily keep all this conversation with the eventually merged PR).

I have only briefly tested this, because I don't have any real testcases where `DILUTE` is not set to `NOTUSED` (it would be very useful to such a testcase to the behaviour testsuite - please feel free to add one!). So please let me know if this doesn't work for any reason. The only slight wrinkle I see is that it doesn't update the line near the bottom of `mechanism.f90`:
```
!* End of Subset.  No. of Species = 30, No. of Reactions = 71 ;
```
to reflect the new reactions created. But that line is not used anywhere, it's just cruft from the original file, so I don't see an issue. (In the case I used, the count of species is out by one anyway! - not sure why).

rs028:
Hi @spco, for the purpose of updating the manual, is this correct?

- adds DILUTE process for all species in the chemical mechanism, if `DILUTE` in `environmentVariables.config` is NOT set to `NOTUSED`, otherwise leaves the mechanism as is.

- If `DILUTE` is set to zero, the chemical mechanism is modified (i.e., `0` is not the same as `NOTUSED`).

- in principle, `DILUTE` can be constrained.

spco:
1 and 2 are correct, but `DILUTE` cannot be constrained as of now, because the current implementation hard-codes the value from `environmentVariables.config` into `mechanism.f90`. You've highlighted that I could tweak things easily to allow the use of constrained values though - I'll make a change to #412 to reflect that.

spco:
Hi @bsn502 - you'll see that the `full` test fails on both the Linux and Mac builds. This is discussed at https://github.com/AtChem/AtChem2/pull/412#issuecomment-571563440 . @rs028 how would you like to proceed with this? Until it's fixed, Travis will always fail. Options I see are to
(a) tweak `full` to give a problem that can actually be solved rather than stalling;
(b) remove `full` from the testsuite by deleting its folder;
(c) half-remove `full` by leaving the folder there, but tweaking the test runner script to not run `full`;
(d) something else.

(a) is possibly a lot of work, (b) means we lose the big test, (c) means `full` is there for future use if we wanted to reinstate it with a bit of work.

Personally, maybe (b) is the right option for now - we can always manually reinstate `full` from earlier commits if need be, if it will add value - what do you all think?

spco:
Oh, and thanks for all your continuing work @bsn502 !

rs028:
@spco yes I think that is sensible. We want to revamp the testsuite anyways at some point.

@bsn502 Can you delete the `full/` directory in `travis/tests/` please? 

I guess we don't need `dilute.py` anymore, as well. And ` build/build_atchem2.sh` should revert to its original version?

spco:
Sure, I think either solution would be fine - perhaps the acknowledgement line is the right way to go?

spco:
Probably best to `squash and merge` given the convoluted nature of the commit history in this PR :)

rs028:
No need to do a rebase?

rs028:
@bsn502 can you add an acknowledgement line to the header similar to line 15 in `tools/plot/plot-atchem2_v3.py`?

build/mech_converter.py
```diff
@@ -1,3 +1,4 @@
+-------------------------------------------------------------------------
```
spco: Needs a `#` to mark this line as a comment, not a command :)

bsn502: I'm on it :)

rs028:
Is this ready to be merged? @spco does it need a rebase?

spco:
Looks like it to me. I would squash and merge using the GitHub GUI, given the slightly circuitous commit history to what is essentially changes to one file and the removal of a directory.

PR #412, opened 2020-01-06
Testing of dilute fix

spco:
My personal version of #408 for my testing and development purposes

### DIFF
--- a/build/mech_converter.py
+++ b/build/mech_converter.py
@@ -106,7 +106,8 @@ def convert(input_file, mech_dir, mcm_dir):
 
     :param input_file: string containing a relative or absolute reference to the mcm file to be processed.
     :param mech_dir: string containing a relative or absolute reference to the directory in which the function should
-      place mechanism.f90. This is normally model/configuration.
+      place mechanism.f90, and where the environmentVariables.config file should be read from.
+      This is normally model/configuration.
     :param mech_dir: string containing a relative or absolute reference to the directory in which the function should
       place mechanism.{prod,reac,ro2,species}. This is normally model/configuration/ for the given model.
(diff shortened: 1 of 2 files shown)


rs028:
TBH, I am not entirely sure constraining DILUTE is a needed feature. It won't hurt to have that possibility, but I can't imagine many cases were it's needed. So, if it is too much work, I think you can leave it :)

spco:
Seems like my PR was breaking the tests even before I removed the hard-coding, so I'm trying to understand that...

spco:
@rs028 - so it turns out the only issue was that the `full` test has a non-zero DILUTE value, which seems to cause a stall in the solve (and would obviously also have a different solution now that DILUTE is being treated correctly). It may be that we need to change that test to something that actually solves correctly - I've only managed to get this PR to complete by skipping the `full` test.

rs028:
Meaning this has nothing to do with constraining or not, I guess.
Obviously the full test results will need to be updated. But this means that the model cannot complete the run when it is using the dilute value?

spco:
Yeah - up until now the dilute value hasn't been used, so we're just changing the solution (and evidently that hits a solvability issue). I think that's purely a function of this specific model (`full`), so we may just be able to tweak it until it solves. Perhaps the dilution factor (about 1e-6) is too small - that could well cause numerical issues

rs028:
I think that may well be the case

rs028:
I think we are done with this, no? `DILUTE` cannot be constrained, but that is fine. Very few applications require that and we know how to implement that, if needed.

spco:
It would seem so from my perspective!

rs028:
@spco this could be closed, or do you need it for further testing?

spco:
Thanks all!

Issue #413, opened 2020-01-16
Reactions with no reactants


mpanagi:
When you write reactions with no reactants the model doesn't run.

You need to give a reactant name (e.g. EMISSIONS) to run the model

rs028:
To expand on this:

The reaction:

    % k : = A;

creates an error in `mechanism.species`. This is not the case when the products are missing (eg, `% k : A = ;`) which seems to be working fine.

A reaction with no reactants can be used to simulate emissions and one workaround this problem is to create a fake species with a very large constant value and use it as a "reactant". This is not ideal, the python script should be able to deal with missing reactants as it does with missing products.

spco:
Ok - so is it that `mech_converter.py` either errors or creates a malformed `mechanism.species`, or is it that AtChem can't handle it, even though it is perfectly well formed?

rs028:
I think it is a malformed `mechanism.species`. It looks like this:

```
A
B

C
```

and probably the system does not like the empty line. @mpanagi  can you confirm?

rs028:
I have a done a simple test using the following basic mechanism:
```
% 1 : A + B = C ;                                                                             
% 2 : = D ;                                                                                   
% 3 : E + F = G ;                                                                             
% 4 : H = ;
```

The files `mechanism.reac`
```
9 4 0 numberOfSpecies numberOfReactions numberOfGenericComplex
1 1
1 2
2 4
3 6
3 7
4 9
```

 and `mechanism.prod`:
```
9 4 0 numberOfSpecies numberOfReactions numberOfGenericComplex
1 3
2 5
3 8
```

both seem correct. The file `mechanism.species` however is not:
```
1 A
2 B
3 C
4 
5 D
6 E
7 F
8 G
9 H
```
What I don't understand is why reaction 2 causes a problem but reaction 4 does not. Trying to locate the point of `mech_converter.py` where this happens.

spco:
I don't have time to fully test this, but I believe the issue is at https://github.com/AtChem/AtChem2/blob/057fdffe52ad3e95208314a38a64318db1296b02/build/mech_converter.py#L325 
`reactantsList` is ' ' (one space) rather than the empty string, so it's entering the body of the if statement when it shouldn't. It should be fixable with 
```
if not reactantsList == '' and not reactantsList.isspace():
```
or similar, but I haven't fully tested that.

[...]

PR #422, opened 2020-04-25
Prepare release v1.2

rs028:


rs028:
@spco These `.gitignore` files from the `travis/tests` subdirectory should not be needed and can be removed. However this creates a bunch of empty directories when constraints are not used in a test. Can/should these be removed too?

travis/run_tests.sh
```diff
@@ -208,28 +208,28 @@ done
 if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then bash <(curl -s https://codecov.io/bash) -F tests ; fi
 
 # After all tests are run, exit with a FAIL if $fail_counter>0, otherwise PASS.
-if [[ "$style_test_passed" -gt 0 ]]; then
```
spco, line 211: Are these lines meant to be commented out?

spco: Sorry, just seen your edited comment about the tests. I think this is here to provide a summary of all the tests in one place at the end of the testsuite - so if something fails, you can see it here. I found it made the Travis logs a little easier to navigate. It's up to you whether you want to remove it or not.

spco:
On the subject of the empty directories, I can't remember whether they are needed or not - do the tests complete if they're removed? Remember that git only tracks files, not directories, so an empty directory will not be replicate if you clone the repo.

If the directories do need to be kept for the tests to run etc (or just to make it clear to users what the directory structure should be - they can just look in any test directory), then the usual convention is to put a blank `.gitkeep` file inside them, so that git can keep track of them. They're not special files like `.gitignore`, so git doesn't treat them any differently to any other file.

travis/run_tests.sh
rs028: Yes I see that, but `make tests` does not execute the indent and style tests so this always report them as passed which is pointless. It makes sense only for `make alltests` in which case I agree it is useful. Unless I am reading this wrong.

rs028:
The tests pass without the empty directories. The question is, will they remain in the repository since I cannot push their deletion? The rationale for removing them is to simplify the addition of new tests.

spco:
Directories are not tracked if empty, so they will not show up - there is nothing to delete.

Good point about making it easier to add new tests. Is there a test that uses all of the directories? It might be useful to be able to point users to a single canonical example where they're all used, just to make it clear what is there. Or I guess the relevant section in the manual could be used as the guide instead anyway?

travis/run_tests.sh
spco: You're absolutely right.

rs028: I will remove these lines then. But I agree with you that a summary should be printed at the end of the test suite. I will create a new issue about that.

rs028:
Yes there are tests that use all directories. I think the manual is quite clear on how to add tests. In any case at some point we want to remake the whole test suite so that will be the moment to sort out these issues. I don't have time to work on the test suite now and I would like to have a new release fairly soon (btw I send you a bunch of messages on slack about that, don't know if you are checking it these days).

rs028:
Why is this failing on macOS now?

spco:
Have restarted it - the log is empty so it may just be a Travis issue.

spco:
It's failed again in the same way. Nothing in their changelog since 9 days ago, so don't think it can be that. https://www.traviscistatus.com/ is showing no issues, but I think it's at their end. May have to wait a couple of hours and retry.

rs028:
It was failing 2 days ago as well. It's not urgent, and at least it runs on Linux.

rs028:
I think this is ready to go. I will wait a bit to merge while @mpanagi tests the solution to #413.

mpanagi:
After the minor edits the model now compiles and runs and the results are the same.

rs028:
Thanks @mpanagi.

If there is nothing else to add I will merge this and make a note somewhere about the travis tests for macOS.

rs028:
Okay version 1.2 is now ready. Thanks to @spco, @bsn502, @mpanagi for your help and contributions!

Issue #407, opened 2019-10-18
Improve efficiency of 'implicit none' statements


spco:
So long as a function is within a module, it's sufficient to have a single `implicit none` statement at the top of the module. So we could remove all the many `implicit none`s from the functions, and have only one per module plus one for the main program, and any function that isn't within a module.

Would likely save us a fair few lines (grep finds 122 occurrences, so we'd probably save about 100 lines)

rs028:
Done with PR #424.

PR #424, opened 2020-04-29
Pre-release v1.2

rs028:
This PR includes a few small improvements and bug fixes before the release of 1.2

### DIFF
--- a/build/fix_mechanism_fac.py
+++ b/build/fix_mechanism_fac.py
@@ -9,8 +9,13 @@
 #
 # -----------------------------------------------------------------------------
 
-# This file contains functions to fix the contents of a .fac file by removing the incorrect newline characters.
-# Call with a single parameter containing the file to fix.
+# This script contains functions to fix the contents of a chemical
+# mechanism file in FACSIMILE format (.fac ) by removing the incorrect
+# newline characters.
+#
+# ARGUMENT:
+# - path to the .fac file
+# ---------------------------------------------- #
 from __future__ import print_function
 from functools import reduce
 import sys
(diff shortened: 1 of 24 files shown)


rs028:
Is this all that is needed to solve #407?

rs028:
Yes I figure alI the fortran files. I started with one to check first that I was not missing something.

rs028:
The script `mech_converter.py` adds `implicit none` to `mechanism.f90` should this be removed as well?

spco:
We should leave it - it could be moved out of `update_p`, and into the module `mechanism_mod`, but up to you whether you move it or leave it where it is.

rs028:
Does this solve #407 completely?

spco:
From what I can grep, that looks right :)

rs028:
@spco I would like to solve at least #413, if possible, before merging this. I imagine you are probably swamped with your stuff these days, but would you be able to point me in the right direction?

rs028:
It seems I was overly optimistic in thinking I might have more time on my hands during the lockdown :) so I think this is a good point to stop and freeze v1.2.

The build still fails on macOS, but this time with an error message:
```
/usr/local/Cellar/gcc@4.9/4.9.4_1/bin/gfortran-4.9  selected as fortran compiler, but this is not a valid filename.
Example usage: ./install_cvode.sh /path/to/install/directory /path/to/fortran/compiler
The command "if [[ "$TRAVIS_OS_NAME" == "osx" ]];   then ./tools/install/install_cvode.sh $PWD /usr/local/Cellar/gcc@4.9/4.9.4_1/bin/gfortran-4.9 ; fi" failed and exited with 1 during .
```

If this is a problem at their end, I will merge it anyways (it is fine on linux).

spco:
Seems we also have a problem to fix - I've cleared the cache and the install of libffi is still failing. It's probably homebrew changing a setting, or Travis changing something internal with their image. Will need to do some digging as to why that is, but not now - I don't see any reason for it to delay v1.2 either.

Issue #427, opened 2020-06-03
Truncation of Rate Definitions and Species Outputs


AlfredMayhew:
Hi, I have run into some problems with the truncation of lines in AtChem2. The two instances are described below:

- A limit of ~125 characters on rates defined under the "Generic Rate Coefficients" and "Complex reactions", a similar limit applies when the rates are defined lower in the file, before a reaction definition. Clearly 125 characters is quite large, but I have found myself wanting to use definitions longer than this and have to split the definition into smaller chunks, which makes the mechanism file quite long.

- More significant is the limit of ~10 characters on species names in output files. I think the names are carried through the calculations fine, it is just when they are outputted that the end of the species name gets trimmed off. This leads to two species such as "ABCDEFGHIJONE" and "ABCDEFGHIJTWO" both being represented in the output files as "ABCDEFGHIJ".
This is not such an issue if implementing mechanism files straight from the MCM (as the species generally have quite short names), but I have found it quite easy to reach this 10 character limit.

I would be grateful if somebody could point me to the code setting these limits so I could increase them for my own purposes, even if it is not deemed necessary for an update (files from the MCM do not run into these issues).

Thanks,
Alfie

spco:
Hi Alfie, nice to meet you.

On the first point, could you provide more detail? What errors do you get, and at what stage? During the python conversion, compilation, or at runtime? I'm not aware of any specific line-length limit (though my memory may fail me here).
It may be that you're hitting your Fortran compiler's line-length limit, so if you can provide your compiler version that might be useful.

On the second point, this is controlled by `maxSpecLength` as defined at https://github.com/AtChem/AtChem2/blob/d01a5d904513aa36c8b023de15d70bdd41033cbb/src/dataStructures.f90#L74 . That value should be OK to lift without incident - please let me know if it does trigger any incorrect behaviour.

Sam

spco:
Hi Alfie,

Ok, the printing itself is being controlled by the output format - see https://github.com/AtChem/AtChem2/blob/d01a5d904513aa36c8b023de15d70bdd41033cbb/src/outputFunctions.f90#L359 for example - '100A15' means it can print up to 100 columns of 15-character strings. So if you increase that (and look through that file to see other places where that's used) that should fix your issue.

[...]

PR #428, opened 2020-06-05
Truncation changes

AlfredMayhew:
Hi,
This request relates to the [recent issue I raised](https://github.com/AtChem/AtChem2/issues/427), and takes two parts (two commits).

I was running into issues with the maximum size of species names allowed. I have increased the maximum species name length, and the maximum length in the output files (concentration and rate outputs), to 50 characters. Clearly this limit is arbitrary, and with the fixed width of columns in the output file this value does alter the readability of the speciesConcentrations.output file. I also increased the grouping of species at the top of the speciesConcentrations.output file from 100 species before a newline character to 1000. Again, this is subjective as a limit but I think that over 100 compounds isn't a ridiculous amount, and it makes processing the output file more difficult to have groupings into 100 species.

Secondly, I added the -ffree-line-length-none flag to the Makefile.skel file to prevent lines being truncated if they are too long in the mechanism file. I ran into this issue because my rate expressions were too long. I added this flag to my Makefile directly previously so have not tested using the .skel file when building a new Makefile, but I'm assuming this should pose no issue.

Thanks,
Alfie

### DIFF
--- a/src/outputFunctions.f90
+++ b/src/outputFunctions.f90
@@ -294,7 +294,7 @@ contains
           ! r contains the occurences of each of the detailed species in reactions.
           ! r should have row lengths as in arrayLen, so all accesss to r(i,j) should
           ! be valid.
-          write (output_file_number, '(ES15.6E3, I14, A12, I15, ES15.6E3, A, A)') t, rSpecies(i), &
+          write (output_file_number, '(ES15.6E3, I14, A50, I15, ES15.6E3, A, A)') t, rSpecies(i), &
                                                                                   trim( speciesNames(rSpecies(i)) ), &
                                                                                   r(i, j)%reaction, &
                                                                                   r(i, j)%frequency * p(r(i, j)%reaction), &
@@ -356,7 +356,7 @@ contains
     end if
 
     if ( first_time .eqv. .true. ) then
-      write (50, '(100A15) ') 't', (trim( specOutReqNames(i) ), i = 1, size( specOutReqNames ))
+      write (50, '(1000A50) ') 't', (trim( specOutReqNames(i) ), i = 1, size( specOutReqNames ))
       first_time = .false.
     end if
 
@@ -365,7 +365,7 @@ contains
         arrayOfConcs(i) = 0.0_DP
       end if
     end do
-    write (50, '(100 (ES15.6E3)) ') t, (arrayOfConcs(i), i = 1, size( arrayOfConcs ))
+    write (50, '(1000 (ES50.6E3)) ') t, (arrayOfConcs(i), i = 1, size( arrayOfConcs ))
 
     return
   end subroutine outputSpeciesOfInterest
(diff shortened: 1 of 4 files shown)


src/outputFunctions.f90
```diff
@@ -294,7 +294,7 @@ subroutine outputRates( rSpecies, r, arrayLen, t, p, flag )
           ! r contains the occurences of each of the detailed species in reactions.
           ! r should have row lengths as in arrayLen, so all accesss to r(i,j) should
           ! be valid.
-          write (output_file_number, '(ES15.6E3, I14, A12, I15, ES15.6E3, A, A)') t, rSpecies(i), &
+          write (output_file_number, '(ES15.6E3, I14, A50, I15, ES15.6E3, A, A)') t, rSpecies(i), &
```
spco, line 297: Could you please amend to 52, in order to ensure there's always whitespace between columns, even when the full 50 characters are used.

src/outputFunctions.f90
```diff
@@ -356,7 +356,7 @@ subroutine outputSpeciesOfInterest( t, specOutReqNames, allSpeciesConcs )
     end if
 
     if ( first_time .eqv. .true. ) then
-      write (50, '(100A15) ') 't', (trim( specOutReqNames(i) ), i = 1, size( specOutReqNames ))
+      write (50, '(1000A50) ') 't', (trim( specOutReqNames(i) ), i = 1, size( specOutReqNames ))
```
spco, line 359: Here, and in line 365, could you experiment with removing the repeat number? From memory, and googling, I _think_ this can be just `'(A15) '`, or possibly it needs to be `'((A15)) '`. And the below would be `'(ES15.6E3) '` or `'((ES15.6E3)) '`. If we can remove the repeat number then we of course have can support arbitrary columns.

Apologies that I don't have a working AtChem setup on this laptop to test it quickly, and my usual Fortran reference book is stuck in my office, and I'm not sure how long it will be before I get to lay hands on it again!

AlfredMayhew: I've had a bit of a play with this. I think removing the repeat number ends up just adding a newline after each entry. Removing it from line 359 adds newline characters between each species name in the header whereas removing from line 365 adds newlines between each value.

spco:
I've just opened #429 to attempt to fix the broken MacOS build, but I won't wait for that to hold this one up. Are you satisfied this is complete, @AlfredMayhew ? Looks it to me, in which case I will merge.

rs028:
Thanks for this guys.

@AlfredMayhew you can also add your name to the `doc/latex/Credits.tex` (line 60) :)

AlfredMayhew:
Thanks very much, I have added the credit (in the wrong place initially!). Hopefully that is all correct.
I'm happy for you to merge the changes, thank you again.

rs028:
@spco, does it need to be rebased to #429 before merging?

spco:
No, this should be fine without rebase.

spco:
Thanks @AlfredMayhew!

Issue #17, opened 2017-03-21
Add new environmental constraints


rs028:
A procedure to add new environmental constraints (besides TEMP, PRESSURE, etc) by simply putting the file in the environmentalConstraints directory and adding the corresponding line to modelConfiguration/environmentVariables.config

spco:
#157 has worked towards this, it's still not there yet though.

rs028:
#432 show the procedure to add a new variable manually.

PR #432, opened 2020-07-10
New environment variable

rs028:
Create a new environment variable `ASA` (aerosol surface area). It should work in the same way as `BLHEIGHT`, e.g., can be `NOTUSED` (default), `CONSTRAINED` or given a fixed value, and it can be used to calculate rate coefficients. For example:

    % ASA*525.068*(TEMP@0.5)*2.0D-3   :   O3   =   ;


rs028:
@spco Apart from the tests failing, which I will fix later, is this all there is to do to add a new environment variable?

spco:
Looks about so from searching the codebase. You'll also need to update https://github.com/AtChem/AtChem2/blob/018cf029d879f0f353bc64d7aada9844f9ef3deb/build/mech_converter.py#L432 to match new signature, and add `ASA` to reservedSpeciesList at https://github.com/AtChem/AtChem2/blob/018cf029d879f0f353bc64d7aada9844f9ef3deb/build/mech_converter.py#L37 . I can't guarantee that that's everything, but looks roughly correct.

spco:
Looks about so from searching the codebase. You'll also need to update https://github.com/AtChem/AtChem2/blob/018cf029d879f0f353bc64d7aada9844f9ef3deb/build/mech_converter.py#L432 to match new signature, and add `ASA` to reservedSpeciesList at https://github.com/AtChem/AtChem2/blob/018cf029d879f0f353bc64d7aada9844f9ef3deb/build/mech_converter.py#L37 . I can't guarantee that that's everything, but looks roughly correct.

src/constraintFunctions.f90
```diff
@@ -219,15 +219,15 @@ subroutine getEnvVarsAtT( t )
...
     ! To add another environment variable, the user would need to add
     ! that line to environmentVariables.config, and then add this as
-    ! element 11 in the orderedEnvVarNames initialisation below.  Its
+    ! element 12 in the orderedEnvVarNames initialisation below.  Its
     ! treatment needs defining in each of cases 1-3 and default below.
 
-    if ( size( envVarNames ) /= 10 ) then
-      write(stderr,*) 'size( envVarNames ) /= 10 in getEnvVarsAtT().'
+    if ( size( envVarNames ) /= 11 ) then
+      write(stderr,*) 'size( envVarNames ) /= 11 in getEnvVarsAtT().'
```
rs028, line 230: I keep getting an error about the size of `orderedEnvVarNames`. This is the same as the size of `envVarNames` but I can't find where the size of  `envVarNames`  is set.

src/constraintFunctions.f90
```diff
@@ -239,6 +239,7 @@ subroutine getEnvVarsAtT( t )
     orderedEnvVarNames(8) = 'JFAC'
```
rs028, line 239: Also, does it matter that the index of these varables is not the same number as in the `environmentVariables.config` file?

src/constraintFunctions.f90
spco: `envVarNames` is allocated at https://github.com/AtChem/AtChem2/blob/e18a0341a11b4a9f7d57e46c2fd0a7c01cc18d85/src/inputFunctions.f90#L794 , which is based on  the number of lines in the .config file at https://github.com/AtChem/AtChem2/blob/e18a0341a11b4a9f7d57e46c2fd0a7c01cc18d85/src/inputFunctions.f90#L788 (read the comment for more info).

src/constraintFunctions.f90
spco: No, it doesn't matter - if I recall correctly, the order here is defined based on which env vars can depend on others (so those that may depend on others are calculated later in the list). It's independent of the order in the input file.

src/constraintFunctions.f90
spco: I guess we're going to run into trouble here because the new `ASA` is optional. In the existing setup, while the order of env vars doesn't matter, they all must be present in `environmentVariables.config`. So in the current setup, you'd need to add a line for `ASA` to each of the `environmentVariables.config` files for each of the tests. That of course is not a very robust way of doing things.

We'd likely want to move to a position where `ASA` is optional - that would need some work to make the existing processing more flexible.

I think that's perfectly possible, but will need some work - this current setup is largely a legacy of the direct porting from the original.

rs028: Okay, it seems that this would be the objective of issue #17, whenever we get to it.

src/constraintFunctions.f90
rs028: All right. I guess the index in `environmentVariables.config` is not really necessary anyway. This can be tidied up as part of #17, as well.

travis/tests/firstorder/firstorder.out.cmp
```diff
@@ -31,16 +31,16 @@ AtChem2 v1.3-dev
  Finished reading species names.
 
  Reading initial concentrations...
-      1 A                1.000E+02
-      2 B                0.000E+00
+      1 A                                                        1.000E+02
```
rs028, line 34: Not that it matters, but what is going on with all these extra spaces in the `*.out` files?

spco: I'd imagine it's the text editor you're using doing some automatic conversion, likely something about how many spaces are equivalent to 1 tab.

rs028: Must be something like that because I don't think I changed any thing that do that. One of these days I am going to get rid of all tabs in the code :)

rs028:
If there are no more comments I think this can be merged.

wangshenbo0701 (commented):
I want to add the uptake coefficient (SQX) to the environment variables. Following the ASA method you added, I modified mech_converter.py, environmentVariables.config, constraintFunctions.f90, inputFunctions.f90, and solverFunctions.f90, appending 'SQX' after the corresponding 'ASA' code in each file. However, I still encountered the error: "readEnvVar(): Invalid environment variable SQX given in environmentVariables.config." How can I troubleshoot this issue?

rs028:
Hi @wangshenbo0701 First of all, which version of the model are you using? Some of the files  you mention have been modified recently, so the procedure described here may be a bit out-of the date. I can't be sure without looking at the code, but I would start debugging by looking at all the files which contain mentions of environment variables.

Secondly, think whether creating a new environment variable is really needed. It depends on what you are trying to do, of course, but there is also the possibility of writing a custom Fortran function in `customRateFuncs.f90` (check the manual for details).

Issue #437, opened 2020-12-18
Stack smashing detected when run the test


ShenAlbert:
ubuntu@VM-0-12-ubuntu:~/AtChem2$ ./atchem2
AtChem2 v1.2

-------------
 Directories
-------------
 Model directory is: model
 Output directory is: model/output
 Reaction Rates directory is: model/output/reactionRates
 Configuration directory is: model/configuration
 Constraints directory is: model/constraints
 Environment Constraints directory is: model/constraints/environment
 Photolysis Constraints directory is: model/constraints/photolysis
 Species Constraints directory is: model/constraints/species
 MCM directory is: mcm
 Shared library is: model/configuration/mechanism.so

-----------------------
 Species and reactions
-----------------------
 Number of Species   = 29
 Number of Reactions = 71

 Size of lhs = 114
 Size of rhs = 100

 Reading reactants (lhs) from mechanism.reac...
 Reading products (rhs) from mechanism.prod...
 Finished reading lhs and rhs data.

 Reading species names from mechanism.species...
 Finished reading species names.

 Reading initial concentrations...
      1 CH4              4.900E+13
 ...
      4 NO2              2.400E+11
 Finished reading initial concentrations.

----------------------------------------
 Species requiring detailed rate output
----------------------------------------
 Reading which species require detailed rate output...
      1 OH        
      2 HO2       
 Finished reading which species require detailed rate output.
 Species requiring detailed rate output (number of species found): 2

 Reading ro2 numbers from mechanism.ro2...
 Finished reading ro2 numbers.
 Reading solver parameters from file...
 ------------------
 Solver parameters:
 ------------------
            atol:   1.000E-03
            rtol:   1.000E-04
       deltaMain:   1.000E-04
        lookBack:         100
         maxStep:   1.000E+02
 preconBandUpper:         750
 preconBandLower:         750
      solverType: SPGMR + Banded Preconditioner 
 ------------------

 Finished reading solver parameters from file.

 Reading model parameters from file...
 -----------------
 Model parameters:
 -----------------
                                   number of steps:          48
                                         step size:   0.900E+03
                      species interpolation method: piecewise linear 
                   conditions interpolation method: piecewise linear 
                 rates output step size, ROPA/RODA:        3600
                                  model start time:           0
                         jacobian output step size:           0
                                          latitude:   0.515E+02
                                         longitude:   0.130E+00
                   reaction rates output step size:        1800
                                    day/month/year:  21/ 6/2010
 -----------------

[...]

Issue #435, opened 2020-10-26
Fortran runtime error


jjunum:
I installed atchem2 and there was no problem for a while.
Suddenly, I believe after some updates in Ubuntu 18.04 LST, I got Fortran runtime errors as follows:

jch@Aerosol:~/Research/AtChem2$ ./atchem2
AtChem2 v1.2-dev

-------------
 Directories
-------------
 Model directory is: model
 Output directory is: model/output
 Reaction Rates directory is: model/output/reactionRates
 Configuration directory is: model/configuration
 Constraints directory is: model/constraints
 Environment Constraints directory is: model/constraints/environment
 Photolysis Constraints directory is: model/constraints/photolysis
 Species Constraints directory is: model/constraints/species
 MCM directory is: mcm
 Shared library is: model/configuration/mechanism.so

-----------------------
 Species and reactions
-----------------------
 Number of Species   = 63
 Number of Reactions = 146

 Size of lhs = 233
 Size of rhs = 204

 Reading reactants (lhs) from mechanism.reac...
 Reading products (rhs) from mechanism.prod...
 Finished reading lhs and rhs data.

 Reading species names from mechanism.species...
 Finished reading species names.

 Reading initial concentrations...
      1 SA               0.000E+00
 ...
     15 O2M              0.000E+00
 Finished reading initial concentrations.

----------------------------------------
 Species requiring detailed rate output
----------------------------------------
 Reading which species require detailed rate output...
      1 SO2AQ
 ...
      9 SULF_NO2
 Finished reading which species require detailed rate output.
 Species requiring detailed rate output (number of species found): 9

 Reading ro2 numbers from mechanism.ro2...
 Finished reading ro2 numbers.
 Reading solver parameters from file...
 ------------------
 Solver parameters:
 ------------------
            atol:   1.000E-03
            rtol:   1.000E-04
       deltaMain:   1.000E-04
        lookBack:         100
         maxStep:   1.000E+03
 preconBandUpper:         750
 preconBandLower:         750
      solverType: SPGMR + Banded Preconditioner
 ------------------

 Finished reading solver parameters from file.

 Reading model parameters from file...
 -----------------
 Model parameters:
 -----------------
                                   number of steps:         929
                               step size (seconds):   0.360E+04
                      species interpolation method: piecewise linear
                   conditions interpolation method: piecewise linear
                               ratesOutputStepSize:       43200
                   reaction rates output step size:       43200
                                    modelStartTime:        3600
[...]

Issue #431, opened 2020-06-28
Need help about the example mechanism file


xiaoky97:
Hi! I followed the manual chapter 2 to install Atchem and run the example mechanism. When I ran ./atchem2, the folling error happened. Could you please give me some help about this? Thanks a lot.

-----------
 Model run
-----------
At line 243 of file src/solverFunctions.f90
Fortran runtime error: Index '61' of dimension 1 of array 'j' above upper bound of 60

Error termination. Backtrace:
#0  0x7f0a0e6aed0a
#1  0x7f0a0e6af819
#2  0x7f0a0e6afe96
#3  0x56120d15b815
#4  0x56120d16287c
#5  0x56120d16fb13
#6  0x56120d1711ea
#7  0x7f0a0e97fc41
#8  0x7f0a0e97f89f
#9  0x7f0a0e97d8fa
#10  0x56120d170b81
#11  0x56120d16cee3
#12  0x56120d16efeb
#13  0x7f0a0e4c20b2
#14  0x56120d11698d
#15  0xffffffffffffffff

spco:
Hi @xiaoky97 , thanks for your question - hopefully we can get you up and running soon.

Firstly, could you please repeat Step 4 of §2.4, and check that (a) there are no errors from that step, and (b) run step 5 afterwards and confirm that you see the same behaviour. Please send us the full output of both those steps.

After that, can you confirm whether you've made any changes, however minor, to any of the data files, such as the .fac, or any .config files?

Finally, in order to check what's happened, we'd need to please see a copy of your `mechanism_test.fac` file, and your `photolysisConstants.config` and `photolysisConstrained.config`.

Thanks!

xiaoky97:
**_Step 4:_**
* facsimile mechanism file: ./mcm/mechanism_test.fac
* fortran mechanism directory [ default = ./model/configuration/ ]:
* mcm data files directory [ default = ./mcm/ ]:

call mech_converter.py
/home/atchem/atchem/mcm
Running fix_fac_file on /home/atchem/atchem/mcm/mechanism_test.fac
/home/atchem/atchem/mcm/mechanism_test.fac: file read in 255 items
0 corrections made - now removing old
Reading input file
looping over inputted RO2s
adding RO2 to ./model/configuration//mechanism.ro2

make sharedlib
[...]

PR #438, opened 2021-01-06
Fix c-bind bug in update_p()

spco:
Use assumed-size arrays (*) rather than assumed-shape (:) for the c-bind `update_p()` function, as the latter is not standards-compliant and does not offer consistent access to lbound. Older versions of `gfortran` did not raise this as an issue, possibly because they have different minimum standards.

This should close 

Note that upgrading Ubuntu versions also changes the gfortran version: (from https://www.scivision.dev/major-gfortran-changes/)

> Ubuntu 20.04 default: gfortran-9
> Ubuntu 18.04 default: gfortran-7

which likely explains the difference in behaviour when upgrading. I can't find any information about what differences there are in this specific case though - the changelogs don't mention anything that looks relevant.

This should address #431 , #435 , and #437 . Could one or more of @ShenAlbert, @jjunum, @vinthony, @xiaoky97, @AlfredMayhew , or @Ekimmai please test this to see whether it fixes their issue?

### DIFF
--- a/build/mech_converter.py
+++ b/build/mech_converter.py
@@ -432,8 +432,8 @@ contains
     subroutine update_p(p, q, TEMP, N2, O2, M, RH, H2O, BLHEIGHT, DEC, JFAC, DILUTE, ROOFOPEN, ASA, J, RO2) bind(c,name='update_p')
 
         integer, parameter :: DP = selected_real_kind( p = 15, r = 307 )
-           real(c_double), intent(inout) :: p(:), q(:)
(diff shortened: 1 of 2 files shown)


AlfredMayhew:
I've done a fresh install with this branch and it all seems to work. Thanks very much.

spco:
Fantastic, thanks @AlfredMayhew . I will merge this. Hope that's useful for you going forwards 👍

ShenAlbert:
Thank you @spco . I have also done a fresh install with this branch, and the test mechanism ran successfully. 
I will use observation data to test the model recently, and I hope to continue communicating with you if there are any problems.
Thank you very much for your help.

PR #444, opened 2021-02-01
Fix missing code coverage files

spco:
Use FORTC in Makefile.skel to test for the fortran compiler family (rather than the exact name of the executable). This should ensure code coverage is tracked on ubuntu Actions runs.


spco:
Frankly, I have no idea what's going on with this - the codecov results are much lower than previously, and I think it's because the unittests are not being correctly counted. But I've no idea what I'm doing wrong, and have spent too long it already!

I will make final commits this so it is at least up and running, but the coverage will be much lower than we saw before.

PR #454, opened 2021-08-02
Update CVODE download URL

kilicomu:
The CVODE releases have moved to GitHub - download URL has been updated and call to `wget` has been modified to save the archive to the filename that matches the rest of the installation script.

PR #455, opened 2021-08-02
Remove '8' from gfortran versions in ci.yml

spco:


spco:
I've merged on the assumption this will work OK - needs #454 to fix the CVODE download in addition.

spco:
#456 fixes this issue properly

PR #457, opened 2021-08-03
Catch missing RO2 species error when prepping

### LINKED ISSUE #453: Array indexing error via mechanism.ro2
Hi!

The `mech_converter.py` script writes its error messages into the files that it is operating on, e.g. the else clause in the following snippet:

https://github.com/AtChem/AtChem2/blob/37503aefb02bb54fa3b04899c68d15afa8b1c8c0/build/mech_converter.py#L450-L466

This results in lines such as the following appearing in the `mechanism.ro2` file:

```
0 ! error RO2 not in mechanism: CHOCOCH2O2 DIETETO2
```

When `mechanism.ro2` is read in through `readRO2species` from `inputFunctions.f90`, this 0 value is stored in the `ro2Numbers` array:

https://github.com/AtChem/AtChem2/blob/37503aefb02bb54fa3b04899c68d15afa8b1c8c0/src/inputFunctions.f90#L1245-L1276

Which propagates through to the `ro2sum` function from `outputFunctions.f90`:

https://github.com/AtChem/AtChem2/blob/37503aefb02bb54fa3b04899c68d15afa8b1c8c0/src/outputFunctions.f90#L25-L42

This can result in an attempt to reference element 0 of the array `y`, which raises an array indexing error at runtime and crashes the model uninformatively.

Is the error message going into the `mechanism.ro2` file by design to cause the model to crash on erroneous `mech_converter.py` output? If so, I think a more informative error message where the crash would happen would help, e.g.

```
if ( min(ro2Numbers) < lbound(y, 1) ) error stop "Bad RO2 numbers - check mechanism.ro2 for errors!"
```

If it's not intentional, it's probably worth raising an exception in the `mech_converter.py` script at the time when the error is encountered so as to prevent the 0 ever getting into the `ro2Numbers` array and forcing the user to correct the issue before running the model.

I guess the other possibility is that it isn't a deal breaking error, in which case I think the `mech_generator.py` script should make a clear warning to the user about the missing RO2 species and the 0 RO2 number shouldn't get into the `ro2Numbers` array.

What do you think? Happy to implement resulting changes, just wanted to make sure I understand any reasoning behind the error messaging before opening a PR.

Killian

kilicomu:
Adds an exception where an RO2 species is encountered in the RO2 sum of the mechanism file but isn't encountered in the mechanism species list.

We shouldn't get to a point where there is an RO2 species in mechanism.ro2 that isn't in the full species list in the mechanism, so I think the user needs to correct their mechanism before building and running the model, as opposed to a warning being shown and them being allowed to erroneously continue.

I had a look at adding some error handling to the `ro2Sum` function in `outputFunctions.f90` - it is a `pure` function which disallows `error stop`. I think there would be a bit too much unpicking to make that error handling neat, so I avoided it.

Let me know what you think - happy to make changes as requested.

Closes #453

### DIFF
--- a/build/mech_converter.py
+++ b/build/mech_converter.py
@@ -463,7 +463,15 @@ end module mechanism_mod
             # This code only executes if the break is NOT called, i.e. if the loop runs to completion without the RO2 being
             # found in the species list
             else:
-                ro2_file.write('0 ! error RO2 not in mechanism: ' + ro2List_i + '\n')
+                error_message = ' ****** Warning: RO2 species "' + str(ro2List_i.strip()) + '" NOT found in the mechanism. Please check the RO2 section of your mechansm file for incorrect species names! ******'
+                error_message = ''.join([
+                  ' ****** ',
+                  'Error: RO2 species "',
+                  str(ro2List_i.strip()),
+                  '" NOT found in the mechanism. Please check the RO2 section',
+                  ' of your mechanism file for incorrect species names!',
+                  ' ******'])
+                raise RuntimeError(error_message)
 
 
 ## ------------------------------------------------------------------ ##
@@ -492,4 +500,8 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    try:
+        main()
+    except RuntimeError as e:
+        print(str(e))
+        sys.exit(os.EX_DATAERR)
--- a/build/mech_converter.py
+++ b/build/mech_converter.py
@@ -463,7 +463,6 @@ end module mechanism_mod
             # This code only executes if the break is NOT called, i.e. if the loop runs to completion without the RO2 being
             # found in the species list
             else:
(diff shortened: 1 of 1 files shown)


build/mech_converter.py
```diff
@@ -463,7 +463,15 @@ def convert(input_file, mech_dir, mcm_dir):
             # This code only executes if the break is NOT called, i.e. if the loop runs to completion without the RO2 being
             # found in the species list
             else:
-                ro2_file.write('0 ! error RO2 not in mechanism: ' + ro2List_i + '\n')
+                error_message = ' ****** Warning: RO2 species "' + str(ro2List_i.strip()) + '" NOT found in the mechanism. Please check the RO2 section of your mechansm file for incorrect species names! ******'
```
spco, line 466: I think this line has snuck into the commit and needs removing.

spco (commented):
Generally looks good (see comment).

build/mech_converter.py
kilicomu: Yep you're right - relic from messing around.

I've removed that line.

PR #459, opened 2021-08-17
Escape colon character in LAPACK_LIBS string

kilicomu:
The colon character is getting passed through to a `make` target unescaped, which causes the test for LAPACK libraries to fail.


kilicomu:
Initially, I'm not sure how this change will affect the CI builds, so I'm opening in draft to see what happens when the LAPACK libs are visible to SUNDIALS.

kilicomu:
Doesn't look like escaping the colon has led to functional LAPACK. In my head this means that the build couldn't find the LAPACK libraries at the specified location, so I need to poke around in the build.

kilicomu:
LAPACK is now functional in the CVODE build, but we're in the same position that I am on my system - undefined symbols in the CVODE libraries which propagate through to the AtChem2 build:

```
/usr/bin/ld: cvode/lib/libsundials_cvode.so: undefined reference to `dgbtrs_'
/usr/bin/ld: cvode/lib/libsundials_cvode.so: undefined reference to `dcopy_'
/usr/bin/ld: cvode/lib/libsundials_cvode.so: undefined reference to `dscal_'
/usr/bin/ld: cvode/lib/libsundials_cvode.so: undefined reference to `dgetrf_'
/usr/bin/ld: cvode/lib/libsundials_cvode.so: undefined reference to `dgbtrf_'
/usr/bin/ld: cvode/lib/libsundials_cvode.so: undefined reference to `dgetrs_'
```

(https://github.com/AtChem/AtChem2/pull/459/checks?check_run_id=3349186758#step:12:40)

I guess this means that CVODE isn't correctly linking to OpenBLAS, as these symbols are found in the OpenBLAS library (at least on my system - worth checking in the CI build...).

kilicomu:
The symbols are definitely present in OpenBLAS lib: https://github.com/AtChem/AtChem2/pull/459/checks?check_run_id=3349428841#step:7:1

So, need to look at the SUNDIALS build to see what's going wrong.

PR #464, opened 2022-01-04
New Testsuite

rs028:
Starting work on the new test suite. This PR contains some reorganization of the directories to allow the new and the old tests to run in parallel until work is completed.

To make things clearer, the command to run the build and behavioural tests has been changed from `make tests` to `make modeltests`.


rs028:
@spco Am I reading this wrong or CodeCov is working again? I haven't touched it.

spco:
No idea - as far as I am concerned, CodeCov's input here should just be ignored regardless! :D

rs028:
You mean ignore it for this type of PR?

spco:
Frankly, I don't trust it to be meaningful and/or correct on any PR!

rs028:
Fair enough. Should we consider removing it altogether or mark it as something to be (eventually) fixed?

rs028:
Any idea why test fails on MacOS with gfortran 8? It gives an error about sundials, but it appears that cvode is installed correctly.

spco:
I'm not sure, but I think the error in the `install cvode (MacOS)` step may be something to do with it - it fail to link a simple fortran program.
```
-- Trying to compile and link a simple Fortran program... FAILED
WARNING: Fortran compiler not functional.
   FCMIX support will not be provided.
```

It looks like the image we use - `macos-latest` - has upgraded from 10.15 to 11 recently - see https://github.com/actions/virtual-environments/issues/4060 .

I suspect this may have something to do with it - perhaps v11 has some different permissions, or something else.

To diagnose whether this is the cause, perhaps you could change `macos-latest` at https://github.com/AtChem/AtChem2/blob/4d72237f2abbe72eb26a8cedb1575235f2d7739f/.github/workflows/ci.yml#L25 to `macOS-10.15` (note the capitalisation) and see whether that runs successfully?

rs028:
Looks like that. I will leave it to version 10.15 for the time being and make a note of it.
If there are no further comments I am ready to merge this and start developing the tests.

PR #465, opened 2022-01-05
New Testsuite (2)

rs028:
First tests for the new testsuite ~(do not run when `make tests` is executed)~.


rs028:
While implementing the new tests, I noticed that there are upgraded versions of openlibm and numdiff, so I have included that here.
It seems to fail the tests on mac, but I am not sure if it is a caused by the upgrade or something else).

rs028:
When this is okay, I will merge it and add more tests in a separate PR, as this one already has a lot of changes.

.github/workflows/ci.yml
```diff
@@ -74,7 +74,7 @@ jobs:
           cp tools/install/Makefile.skel Makefile
           ./build/build_atchem2.sh ./mcm/mechanism_test.fac
           if [ $RUNNER_OS != 'Linux' ] ; then
-            install_name_tool -change libopenlibm.dylib.1 $PWD/openlibm-0.4.1/libopenlibm.dylib.1 atchem2 ;
+            install_name_tool -change libopenlibm.dylib.1 $PWD/openlibm-0.8.1/libopenlibm.dylib.1 atchem2 ;
```
spco, line 77: Try changing to 
```
install_name_tool -change libopenlibm.dylib.1 $PWD/openlibm-0.8.1/libopenlibm.4.dylib atchem2 ;
```
(I'm not certain this is the right fix, I'll have another stab if it's wrong)

spco: If not, try 
```
install_name_tool -change libopenlibm.4.dylib $PWD/openlibm-0.8.1/libopenlibm.4.dylib atchem2 ;
```

rs028: Great thanks!

rs028:
Now the original behavioural tests are renamed `oldtests` and the newly developed tests are being added to `modeltests`. Both sets of tests are running under the CI setup.

Eventually the `oldtests` will be deleted (if obsolete or redundant) and/or moved to `modeltests`.

Issue #372, opened 2019-01-09
Format of photolysisConstant


spco:
From #371: Rather than the format
```
1   1e-3   J1
```
`photolysisConstant.config` should use the format
```
J1   1e-3
```
and recreate the photo number from the name (removing the J). TODO: see whether similar can be done for `photolysisConstrained.config`.

rs028:
Doesn't  `photolysisConstrained.config` already works like that?

spco:
Yes, looks like it's just a list of the rate names.

xcy12:
I may also have confusion about J values. J61 actually does not occur in the mechanism.fac file. But the PhotolysisOutput gives J61. I wonder what is J61 used for. Thank you in advance and best regards.

rs028:
@xcy12 This has to do with the MCM. I think that either there was a reaction in the MCM using J61 which was then removed or changed, or it is a placeholder for future work. You can ignore it.

xcy12:
> @xcy12 This has to do with the MCM. I think that either there was a reaction in the MCM using J61 which was then removed or changed, or it is a placeholder for future work. You can ignore it.

Got it! Thank you for your kind explanation.

AlfredMayhew:
Hi,
I'm having issues trying to use the photolysisConstant.config file. I've looked at some of the other issues around this (including this one and #371 ) and I'm just wondering if this was ever resolved. I've noted what I've tried below.

- Initially, I tried formatting as is listed in the manual:
 `1 1E-5 J1`
but this raised the error:
`Finished reading photolysis numbers.
 Number of photolysis numbers: 35
 Looking for photolysis constants file...
 Checking that photolysis constants exist in file...
 Reading photolysis constants from file...
STOP readPhotolysisConstants(): error reading file`
-Then I saw this issue and changed the config file to read:
`J1 1E-5`
but got the same error.
- Then I noticed the point about the ignoring of the first line in #376 so tried adding a blank line at the start of the file. This made no difference when using the `J1 1E-5` format but when I used the  `1 1E-5 J1` format I got a different error:
```
Finished reading photolysis numbers.
 Number of photolysis numbers: 35
 Looking for photolysis constants file...
 Checking that photolysis constants exist in file...
 Reading photolysis constants from file...
At line 357 of file src/inputFunctions.f90 (unit = 6, file = 'stdout')
Fortran runtime error: Expected REAL for item 3 in formatted transfer, got CHARACTER
[...]

PR #467, opened 2022-02-01
Fixed input and output of constant photolysis rates

AlfredMayhew:
Following discussion in #372.

I have also removed the ignoring of the first line in the `photolysisConstant.config` file, when read by `inputFunctions.f90`.

Changing the `inputFunctions.f90` file as discussed in #372 resulted in issues with the `outputFunctions.f90` file, which I believe I have also fixed. 
It's worth noting that when using at least one constant J value only the constant J values are output in the `photolysisRates.output` file. I assumed this was by design since the `photolysisConstant.config` overrides other photolysis rates and sets them to 0 (even if they are specified as constrained), e.g. giving a value of 1E-5 for J1 in `photolysisConstant.config` and listing J2 in `photolysisConstrained.config` will result in a model where J1 = 1E-5 throughout and J2 = 0, regardless of what the J2 constraint file says.

Hopefully this is all OK and makes sense, let me know if there's anything else that needs changing.

### DIFF
--- a/src/outputFunctions.f90
+++ b/src/outputFunctions.f90
@@ -128,8 +128,8 @@ contains
   ! Write photolysis rates to file.
   subroutine outputPhotolysisRates( t )
     use types_mod
-    use photolysis_rates_mod, only : ck, constrainedPhotoNumbers, j, PR_type, &
-                                     constantPhotoNames, numConstantPhotoRates, &
+    use photolysis_rates_mod, only : ck, constrainedPhotoNumbers, constantPhotoNumbers, &
+                                     j, PR_type, constantPhotoNames, numConstantPhotoRates, &
                                      constrainedPhotoNames, numConstrainedPhotoRates, &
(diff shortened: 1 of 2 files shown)


rs028:
Are we happy to merge this? Looks all right to me.

src/outputFunctions.f90
```diff
@@ -145,7 +145,7 @@ subroutine outputPhotolysisRates( t )
           write (58, '(100A15) ') 't', (trim( constantPhotoNames(i) ), i = 1_NPI, numConstantPhotoRates)
           firstTime = .false.
         end if
-        write (58, '(100 (ES15.6E3)) ') t, (j(ck(i)), i = 1_NPI, numConstantPhotoRates)
+        write (58, '(100 (ES15.6E3)) ') t, (j(constantPhotoNumbers(i)), i = 1_NPI, numConstantPhotoRates)
```
spco, line 148: Good bug catch, thanks!

spco:
Looks good to me!

PR #481, opened 2022-09-28
Corrected information regarding the constraining of DILUTE

AlfredMayhew:
The manual states that DILUTE can't be constrained, but this isn't true. I've corrected the manual, hopefully this is all OK.

Thanks,
Alfie


rs028:
Nicely spotted, thanks!

Issue #269, opened 2018-04-25
Plotting tools


rs028:
Python script:
- ~import too many libraries?~
- generally slow
- ~inefficient way to read file header and data (open file twice)~
- ~spacing between panels~

gnuplot script:
- ~does not read variable names from file header~
- ~unequal dimension of panels~
- ~spacing between panels~

Matlab/Octave script:
- ~inefficient way to read file header and data (open file twice)~
- ~does not save to pdf~

rs028:
Python version fixed by PR #394, but still slow. The version with pandas is only marginally quicker than the version with numpy. (keep both?)

spco:
For reference, it's not the data processing that's really expensive - it's the many calls to `savefig()` and `tight_layout()` that do the damage, as seen in this flame graph (this is for the case I'm investigating in #384). `save_fig()` is taking 0.7s per PDF page (of which there are 12 = 8.4s), while `tight_layout()` is taking 0.192s per graph (there are 62 of these = 11.9s) :
![Screenshot 2019-05-21 at 14 00 15](https://user-images.githubusercontent.com/11610738/58098221-fbd58800-7bd0-11e9-956e-991a488affbb.png)

PR #490, opened 2023-03-23
Fix plotting tools

rs028:
A little time to kill in between flights... this is meant to resolve the issues in #269.


rs028:
@spco, this is starting to be annoying... in PR  #482, all the macos tests fail. On this PR only one test for macos12 fails (and for a different reason... the codebase is basically the same!  😡

rs028:
I will merge this, even though thte tests fails, since it only changes the plotting tools, which are independent from the rest of the code.

Issue #304, opened 2018-07-13
Mechanisms in other formats


rs028:
The MCM can be downloaded in other formats, besides FACSIMILE:

- KPP format
- XML format (?)

xcy12:
hello，how to add emissions? for example, I want to add vehicle HONO emission in the model. however, HONO emission from the vehicle is thought to occupy 0.3-0.8% if the NOX emission. so I need to add NOX emission to the model.

xcy12:
in addition, I want to add the sedimentation process for some gases. How can I do that?

spco:
Hi @xcy12 - I have opened a new issue for this discussion at #391.

xcy12:
> Hi @xcy12 - I have opened a new issue for this discussion at #391.
It's very nice to have a new issue about this! Thanks a lot! My partners and I will keep on following this issue.

PR #497, opened 2023-05-17
Add support for kpp format

rs028:
This will let the model use chemical mechanisms in KPP format (issue #304).

Also includes minor and cosmetic changes to the Python scripts in `build/` and `tools/` (checked using pylint) and to some bash scripts.

This also changes the version number back to development towards the 1.3 release (`1.3-dev`).


tools/fix_indent.py
```diff
@@ -130,49 +131,49 @@ def even_quotes(string):
...
-                      and not re.match('^\s*module procedure \s*', to_output, flags=re.IGNORECASE)) \
-                  or re.match('^\s*contains\s*', to_output, flags=re.IGNORECASE) \
-                  or re.match('^\s*program\s*', to_output, flags=re.IGNORECASE) \
-                  or re.match('^\s*interface\s*', to_output, flags=re.IGNORECASE) \
-                  or re.match('^\s*abstract interface\s*', to_output, flags=re.IGNORECASE) \
-                  or re.match('^\s*pure function\s*', to_output, flags=re.IGNORECASE) \
-                  or re.match('^\s*select\s*', to_output, flags=re.IGNORECASE) \
-                  or re.match('^\s*case\s*', to_output, flags=re.IGNORECASE) \
-                  or re.match('^\s*type\s+', to_output, flags=re.IGNORECASE):
+                if re.search(r'\s*if.+then', to_output, flags=re.IGNORECASE) \
```
rs028, line 152: I changed this because the `fix_style.py` script changes IF-THEN-ELSE to lowercase, but it shouldn't make a difference given the flag

build/mech_converter.py
```diff
@@ -170,257 +218,287 @@ def convert(input_file, mech_dir, mcm_dir):
...
 
-    # Check each of the RO2s from 'Peroxy radicals' are in the reference RO2 list. If not print a warning at the top of
-    # mechanism.f90 for each errant species.
-    # TODO: This will break the exected format when mechanism.f90 is replaced by a parsable format.
+    # Check that each of the RO2s from 'Peroxy radicals' are present
+    # in the RO2 reference list from the MCM. If not, print a warning
+    # at the top of mechanism.f90 for each errant species.
+    #
+    # TODO: This will break the expected format when mechanism.f90 is
+    #       replaced by a parsable format
```
rs028, line 253: This TODO comment was added with PR #347. I think it refers to the implementation of the function parser which then was abandoned in favor of the shared `.so` file (?)
If so, it's probably not something "to do" anymore?

rs028:
The procedure involves a new script `kpp_conversion.py` which converts the mechanism in KPP format to a `.fac` file, which is then processed as usual. Possibly not the most efficient way to do it, but the easiest to implement.

This only works for the KPP files created by the MCM website. There is a large variability in the possible format of kpp files so it is hard to do better without specific examples.

Two new tests are also added to verify that the conversion is correct: a reference test with a normal fac mechanism and an identical (ie using same config) test using a kpp mechanism. They give the same results.

tools/fix_indent.py
spco: Makes sense

build/mech_converter.py
spco: I think so - fine to delete.

spco (approved):

spco:
I think this all makes sense. On formatting and style, I have thoughts we should use existing external tools rather than rolling our own if possible - will create a separate issue.

rs028:
@spco thanks a lot!

PR #500, opened 2023-08-04
Corrected example mechanism path in README

stulacy:
The example mechanism has been moved to `model/mechanism.fac` but the path hadn't been updated in the README. This PR corrects this.


rs028 (approved):
Thank you!

PR #501, opened 2023-08-23
Code coverage

rs028:


tests/run_model_tests.sh
```diff
@@ -215,7 +215,7 @@ $this_file_failures"
   echo $this_test_failures >> $LOG_FILE
 done
 
-if [[ "$RUNNER_OS" == "Linux" ]]; then bash <(curl -s https://codecov.io/bash) -F tests ; fi
+#if [[ "$RUNNER_OS" == "Linux" ]]; then bash <(curl -s https://codecov.io/bash) -F tests ; fi
```
rs028, line 218: @JamesAllsopp is this needed at all?

rs028:
Still needs some tidying up but it should more or less work. Basically it does 2 things:
1. introduces a new variable in `Makefile`. If set to FALSE (default) adds `-O2` to the gfortran flags. If set to TRUE, adds the `gcov` flags needed by codecov, but not needed for production runs (for which we want optimization instead).
2. reorganize the `ci.yml` file so that there is no longer a standard run that is not part of a test and then it runs the tests twice: the first for numerical and behavioural accuracy (with -O2 flag) and the second for codecoverage (with `gcov` flags). The second run is doen only for the latest linux and gfortran versions and uploads the results to Codecov for analysis.

tools/install/Makefile.skel
```diff
@@ -22,23 +21,21 @@ FORTC = "gnu"
...
-
 # =========================================================================== #
 # DO NOT MODIFY BELOW THIS LINE
 # =========================================================================== #
 
-
 .SUFFIXES:
 .SUFFIXES: .f90 .o
-.PHONY: all test
+.PHONY: all indenttest styletest unittests oldtests modeltests alltests clean
```
rs028, line 38: This is confusing to me: should all the rules that do not generate a file listed here? If os, why is `all` which generates the executable?

tools/install/Makefile.skel
```diff
@@ -143,18 +128,18 @@ styletest:
 	@./tests/run_style_test.sh
 
 unittests: $(fruit_driver)
```
rs028, line 130: I guess these rules (`unittests`, `modeltests`, etc...) could also be moved to the new `makefile.tests` and only the rules `all`, `sharedlib` and `clean` can remain the main `Makefile`

.github/workflows/ci.yml
```diff
@@ -104,31 +108,48 @@ jobs:
...
-          make oldtests
-          #make modeltests #TODO: modeltests are temporarily deactivated (pass on linux, fail on macos)
+          make oldtests CCOV=true
+          #make modeltests CCOV=true
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          #token: ${{ secrets.CODECOV_TOKEN }}
+          #directory: ./tests/unit_tests/
```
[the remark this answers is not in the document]
JamesAllsopp: No I don't think token is needed, just there as they told me to do it! I think it's only relevant for private repos. directory is described here; https://github.com/codecov/codecov-action It's so you definitely can find the coverage reports, but it doesn't look like it's necessary due to the mv command. You definitely want the mv command though, as you'd have to cd to each directory and run this action, which doesn't sound fun. Or run this code for each test with a different directory argument.

tools/install/Makefile.skel
JamesAllsopp: I'm really not sure about this, I'm sorry

tools/install/Makefile.skel
JamesAllsopp: That's what I'd do, one clean path through to compile the executables. That's what I'd want as a user with no interest in the test cases other than they work in ci.yml

tests/run_model_tests.sh
JamesAllsopp: No, just use the codecov-action

JamesAllsopp:
> 

That's really good, should make the code easier to deal with

tools/install/Makefile.skel
spco: Make uses timestamps on files to decide whether to run a a target or skip it - no point in rerunning a target if none of its dependencies have changed. Make typically assumes a given rule creates a file with the same name.

In the case where this is not true (for example, `all`, `test` or indeed any of the `*tests` above), it will always run the target rule, because there's no file there with a timestamp to check.

However, `.PHONY` is provided as a safeguard in case someone inadvertently creates a file with that name - say someone created a file called `modeltests`. `.PHONY` tells make to ignore the presence of the file, and run the rule anyway. So it's just to safeguard against the scenario where someone creates such a file, runs the rule, make checks the timestamp and dependencies, doesn't run the rule, but just reports success anyway.

TL;DR: it's not vital either way, but best to add all of those to `.PHONY` as you've done above.

spco:
I haven't gone through this in any depth, but it looks like a nice improvement! The key thing is - does the coverage now get measured, and change over time, in ways that make sense? :)

tools/install/Makefile.skel
rs028: Thanks for clarifying this. It makes sense. 
So if I now move the `*tests` tules to the new `makefile.tests` I need to split the .PHONY as well? it's not like a variable that gets overwritten when make executes the `include` instruction?

rs028:
> The key thing is - does the coverage now get measured, and change over time, in ways that make sense? :)

That's the question! I think so, but not sure yet. I found out that I can reset the codecov info. So I could remove the tests, reset codecov and add the tests one by one to see what happens. But maybe first I should merge this, to keep things clean.

tools/install/Makefile.skel
spco: https://stackoverflow.com/questions/59294990/how-should-you-use-phony-in-included-makefiles is quite helpful. Essentially, you can do either, but it might make most sense to split it.

rs028: Thanks! I have been scouring stackoverflow for that, but somehow missed that question. Now it all clear.

.github/workflows/ci.yml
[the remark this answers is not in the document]
JamesAllsopp: The problem is due to the way gfortran 11 handles its coverage files, where it writes them into the directory where the src file is. So you could either use the directory option on codecov/codecov-action@v3 to tell it where to find the files but as the files are put in several sub-directories and codecov doesn't search for them, you would have to run this once for each directory. Alternatively, use the mv command to put them all in the root directory, and don't worry about it

JamesAllsopp (approved):

PR #514, opened 2023-12-22
Adding Stoichiometric Coefficients

### LINKED ISSUE #513: Non-integer Stoichiometry
Hi,
I've been working on running AtChem2 with the GEOS-Chem mechanism and have run into an issue that I've come across before, so I thought I'd mention it here in case it was functionality that we could look at adding to AtChem2.

Many lumped chemical mechanisms will include chemical reactions where a non-integer number of products are formed (or less-commonly, a non-integer number of reactants are consumed). This is not true for the MCM due to its semi-explicit nature. For example `A + B --> 0.25C + 0.5D`. 
AtChem2 does not currently support writing reaction stoichiometry in this way, instead relying on duplicating species within the equation (e.g. `A + B --> 2C` is written as `A + B --> C + C`). Again, this works fine for the MCM but I believe it would add a lot more flexibility to AtChem2 to be able to accept numbers before species names as stoichiometry. A search of previous issues brought up [this discussion](https://github.com/AtChem/AtChem2/issues/319#issuecomment-421295265) from 2018 where you agreed that duplicating species was sufficient since it would work for the MCM.

I have got around this issue a couple of different ways in the past by breaking the reactions apart and duplicating the number of species as required. I think the most generalizable way I have found is to split the reaction based on each product, and then add an extra dummy reaction and products to account for the loss of the reactants. For example, `% k : A + B --> 0.25C + 0.5D` could be written as:
```
% k*0.25 : A + B --> C + A + B ;
% k*0.5 : A + B --> D + A + B ;
% k : A + B --> ;
```

This isn't a bad work-around but it does increase the size of the mechanism considerably if there are multiple lumped reactions, so it would be nice to be able to have AtChem2 read the stoichiometry of reactions. This approach would also mess up the rate files output from AtChem2. 
I'm happy to help look into how to implement this if you think it would be doable, I just have no idea where to start looking!

Thanks,
Alfie

AlfredMayhew:
I have looked into adding stoichiometric coefficients following discussions in #513. It turns out there was already a coefficient assigned in `src/inputFunctions.f90` for each species in each reaction, it was just being set to 1 every time. This coefficient of 1 was then already included in calculations in the `resid` subroutine in `src/solverFunctions.f90`. This made it fairly easy to modify the code to have variable coefficients!

I have made the following adjustments:
- Check for stoichiometric coefficients when reading the species (reactants and products) in each reaction in `build/mech_converter.py`. If no coefficient is given then assign a coefficient of 1.
- Carry these coefficients through as a new column in the files `mechanism.reac` and `mechanism.prod`.
- Read the coefficients in  `src/inputFunctions.f90` instead of assigning a value of 1 for all species.

I've run some models to test this and it seems to be behaving well. I've compared against the previous version with species duplicated and the results are identical. I've also tested some edge-cases like setting the coefficient to 0 and non-integer reactant coefficients, which also work as intended.

Let me know if there's any more changes I need to make, otherwise this resolves #513.


rs028:
I haven't tested myself but it looks okay to me. 

A suggestion for a test. Using `model_tests/spec_model_1` as base case, you can change  couple of reactions to use stoichiometric coefficients. for example:
```
HO2 + HO2 = H2O2
GLYOX = CO + CO + H2
```
The results of the test should be identical to the results of `spec_model_1`.

AlfredMayhew:
I added a test of the stoichiometry here, but it is failing. To make the test, I copied the spec_model_1 directory and manually edited the mechanism file along with `mechanism.prod.cmp` and `mechanism.react.cmp` to replace any duplicated species with a single species with a stoichiometric coefficient.

Looking at the files from the test, there only seems to be some small differences in the `reactionRates` files. When comparing the species concentrations, there are no large differences between the two models (the largest difference is 10 molecules cm-3 of HO2, and the largest % difference is 2E-4%), but maybe this falls outside of the tolerance of numdiff?

I don't think it's unreasonable to accept small differences between using the split mechanism and the stoichiometric coefficient mechanism since the problem being passed to the solver is slightly different, but I wanted to check that it would be OK to adjust this test to use the output from my stoichiometric model rather than comparing to the exact output from spec_model_1?

Note that I have also made changes to the .kpp -> .fac conversion ( c34cdaa ) to allow my kpp test to complete as my filepath contains points ('.'), so was causing errors in trying to create the new kpp file.

rs028:
I am satisfied that the differences between `spec_model_1` and `spec_model_stoich` are negligible. This can be merged, thank you!

rs028 (approved):
I am satisfied that the differences between `spec_model_1` and `spec_model_stoich` are negligible. This can be merged, thank you!

PR #515, opened 2024-01-05
DILUTE Error with Latest Stoichiometry Update

AlfredMayhew:
I have noticed that all AtChem2 models now fail if `DILUTE` is not set to `NOTUSED`, following the changes in #514 . This is because the added dilution reactions in the `mechanism.reac` file don't have defined stoichiometric coefficients. I have fixed this by adding a coefficient of 1.


rs028 (approved):
Nice catch!

rs028:
One question, does the stoichiometric coeff need to be  a real number or does this work even if the user set it to an integer?

AlfredMayhew:
The accepted formats in the mechanism file should be either a integer or decimal number, with or without a space. i.e. all of the following should be accepted:
- `2 O3`
- `2.0 O3`
- `2O3`
- `2.0O3`

This is then read by python and converted to a float in the `separate_stoichiometry` function. This means it's always written out to `mechanism.reac` and `mechanism.prod` as a decimal (real?). I haven't actually tried manually editing these files to change the 1.0 to an integer of 1. I suspect that would cause a FORTRAN error because it would be the wrong type, but I'm not sure.

One thing that has come to mind that I will test is I'm not sure how the `.kpp` conversion script will handle this change either, so I'll try and test that at some point!

I can't seem to merge the request, so feel free to do that.

AlfredMayhew:
Ok, looking at the kpp conversion script it seems like it treats the reaction lines as a 'rate section' and a 'reaction section' and just swaps them, without digging into what the individual species are. This means it should carry any stoichiometry through. I think the kpp test would have failed too if there was an issue, but I will double check when I have some time.

PR #516, opened 2024-01-23
Revision of the manual

rs028:
Revision of the AtChem2 Manual.
Add a github actions workflow to compile and upload automatically the Manual PDF file.


.github/workflows/ci.yml
```diff
@@ -27,56 +28,60 @@ on:
...
+    # 5. Recompile AtChem2, run unit and model tests for Codecov
     steps:
 
       # -------------------------------------------------------------
       # (1) Checkout the repository under $GITHUB_WORKSPACE
-      - uses: actions/checkout@v3
+      - name: Checkout repo
+        uses: actions/checkout@v4
         with:
           fetch-depth: 2
```
rs028, line 69: @spco is `with: fetch-depth: 2` necessary here? I can't understand if it does something in the context of this workflow.

spco: It reduces the number of commits fetched down to 2. By default, the full history is fetched, which is unnecessary here and bloats things. So it just speeds things up a bit - particularly helpful when we have a large number of large files that have gone through repeated revisions (like the output files). It'll function fine without this line, if you prefer clarity or succinctness.

rs028: Ah okay. The documentation was a bit confusing to me, but this makes sense. Thanks

PR #521, opened 2025-01-10
Correcting Rate Output for Non-integer Stoichiometry

AlfredMayhew:
I recently noticed that the rate output was incorrect from models using a mechanism with stoichiometric coefficients (as was facilitated by #514). The actual rate of the reaction (as used in calculations) seems to be correct, but the stoichiometry was not accounted for when reporting the rate in the `lossRates.output` and `productionRates.output` files. For example, the production rate of `B` in the reaction `A = B + B` would be output differently from the reaction `A = 2B`, despite the concentrations of B in each model being identical. The production rate of B would be half what it should be in `A = 2B` because the stoichiometry was ignored.

Looking into `src/outputFunctions.f90` it looks like the outputted production/loss rate value was calculated by multiplying by the number of occurrences of a compound in a reaction by the reaction rate. This worked fine for the integer stoichiometry required prior to #514 , but now fails to account for instances where a compound may only appear once in a reaction, but with a stoichiometric coefficient. 

I have added a tracker of stoichiometry to the `reaction_frequency_pair` type and then used this to calculate the correct rate in `src/outputFunctions.f90`. My changes do make `reaction_frequency_pair` a slightly unintuitive name, so we could change this if needed. In fact, I believe that the frequency value in this type was only used for the rate output anyway, so we could maybe remove that.

Let me know if you have any questions.
Thanks,
Alfie

### DIFF
--- a/src/atchem2.f90
+++ b/src/atchem2.f90
@@ -225,7 +225,7 @@ PROGRAM ATCHEM2
   allocate (detailedRatesSpecies(size( detailedRatesSpeciesName )))
   allocate (reacDetailedRatesSpecies(size( detailedRatesSpeciesName ), size( clhs, 2 )))
   allocate (prodDetailedRatesSpecies(size( detailedRatesSpeciesName ), size( crhs, 2 )))
-  invalid_reaction_frequency_pair = reaction_frequency_pair(-1_NPI, 0_NPI)
+  invalid_reaction_frequency_pair = reaction_frequency_pair(-1_NPI, 0_NPI, 0.0_DP)
   reacDetailedRatesSpecies(:,:) = invalid_reaction_frequency_pair
   prodDetailedRatesSpecies(:,:) = invalid_reaction_frequency_pair
   allocate (reacDetailedRatesSpeciesLengths(size( detailedRatesSpeciesName )))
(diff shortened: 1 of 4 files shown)


rs028:
Hi @AlfredMayhew thanks for finding and fixing this bug. what do you mean with:
 "I believe that the frequency value in this type was only used for the rate output anyway, so we could maybe remove that."

rs028:
It seems the style test fail. Are you able to run `make styletest` and check the log file (`tests/styletest.log`) to see what is wrong?

AlfredMayhew:
Ok, I think I've fixed the style test issue but now I'm getting another fail... I'll have a look into this at some point soon.

With regards to your other question: `reaction_frequency_pair` previously had two associated values (variables?), called `reaction` and `frequency`. As part of my changes, I added a third variable called `stoich` (in `src/dataStructures.f90`). Then, in `src/outputFunctions.f90`, I essentially replaced the use of `frequency` with the new `stoich` variable. I _think_ that was the only place that `frequency` is used in the code. So I'm saying that if you wanted, we could remove that as a variable and potentially rename `reaction_frequency_pair` since it isn't as descriptive any more (though it still kind of works because the 'frequency' could just refer to the stoichiometry). Hopefully that makes sense, it's not a big deal either way.

rs028:
It may be an issue with the unit tests. These look at individual functions and it may be that some of your modifications touch one or more of those functions. From the actions log it seems to be `config_test.f90`.

Re the `frequency` variable and function name, I am okay to change as you suggest. But I am not too fussed about that. If you aren't too sure, just add a comment about it where the variable is defined.

AlfredMayhew:
Ok, I think I've sorted all of the issues now. The unit test fail was because I hadn't passed the new stoichiometry argument to `findReactionsWithProductOrReactant`. I have passed an empty `ccoeff` array to the argument in `tests/unit_tests/config_test.f90` which passes the test. This test doesn't actually test the `ccoeff` variable though (I could try and incorperate this into the test, but I would need to spend more timr figuring out how!).

I also updated the rate output files in the stoichiometry test, which I had forgotten about previously.

Finally, I have just added a short comment to `src/dataStructures.f90` to explain why the `stoich` value is there.

rs028:
Great. The test run fine on Linux but fails on mac, only because that version of mac is obsolete. So I think it would be okay to merge.
Codecov complains because some of the lines added are not tested: if you have time to write a simple test for `ccoeff` I can leave this PR open. Otherwise I can merge it and you open a new one whenever you have time to write the test. I am fine either way, I am still rewriting the manual, so I am in no rush ;)

AlfredMayhew:
I might not get around to making the test for a while (I've noted it on my to-do), so it may be best to just merge now if you're happy with it and I'll make a new request later. Thanks!

PR #523, opened 2025-05-12
Prepare version 1.2.3

rs028:
Set the version number for the next release (v1.2.3), and update the `CHANGELOG.md` and `README.md` files.
Also fix some bugs with the CI and code coverage workflows.

We are now using `ruby-setup` to force github actions to use a specific version of Ruby, because otherwise FRUIT won't compile (will revisit this if/when FRUIT gets updated).


rs028:
The FAIL test is just codecov complaining that coverage has decreased by 0.25%, so this can be merged

PR #531, opened 2025-06-06
Setup for species injection

rs028:
This modifies the `mechanism.f90` file to accept time (`t`) as a variable. It is necessary to specify when the injection of one or mre species occurs, but in general allows to write time-dependent functions using `customRateFunctions.f90`.

### DIFF
--- a/build/mech_converter.py
+++ b/build/mech_converter.py
@@ -40,7 +40,7 @@ import kpp_conversion
 
 reservedSpeciesList = ['N2', 'O2', 'M', 'RH', 'H2O', 'BLHEIGHT', 'DEC', 'JFAC',
                        'DILUTE', 'ROOF', 'ASA', 'RO2']
-reservedOtherList = ['EXP', 'TEMP', 'PRESS', 'LOG10', 'T', 'J', 't']
+reservedOtherList = ['EXP', 'LOG10', 'TEMP', 'PRESS', 'J', 't']
 
 # =========================== FUNCTIONS =========================== #
 
--- a/build/mech_converter.py
+++ b/build/mech_converter.py
@@ -40,8 +40,7 @@ import kpp_conversion
 
 reservedSpeciesList = ['N2', 'O2', 'M', 'RH', 'H2O', 'BLHEIGHT', 'DEC', 'JFAC',
                        'DILUTE', 'ROOF', 'ASA', 'RO2']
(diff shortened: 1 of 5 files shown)


AlfredMayhew:
Sounds interesting. Is the idea that you can pass 'time' to a custom function that then sets the injection rate. So if I wanted an NO injection at a rate of 1E5 between time 0 and 300, then I add a new reaction to the mechanism like `injFunc(t) : = NO ;`, then define `injFunc` to set the rate to 1E5 if t is between 0 and 300, otherwise set it to 0? I think that sounds sensible and avoids having dummy species to handle injections. 

A couple of things to consider regarding lining up the injection times with the model time-steps:
- If you had large model time-steps, then you might end up with less/more injection than expected if your model timesteps don't line up with your injection times. For example, if you have 15 minute time steps starting at t=0, and you want to inject NO for 20 minutes between t=0 and t=1200, then your model time steps will be at t=0, 900, 1800, ... . So, depending on how your function works, you might get injection only up until 900s, or too much injection, up until 1800s. This can be avoided by lining up the injection times with timesteps and/or having small time steps so that any over/under injection is small.
- If the function instead accepts a single time step, there is potential that the function might 'miss' the injection time, but I think specifying a time range would be more realistic and also avoid this issue (provided the range is wide enough).

Species injections are quite a common thing people might want to simulate for chamber experiments, so it might be nice if we could include some generalised form of the function in `customFunctions.f90` by default, so that users don't have to write their own fortran functions to do this?

Being able to access time in the mechanism also offers some other potentially interesting uses, like time-varying emissions into the model (e.g. to match ambient emissions), which would previously have been achieved with a constrained dummy species.

rs028:
Yes, that is exactly the idea. And as you say there are many possible applications.

I agree it would be good to have a permanent injection function but, as you say, there are many subtleties involved. So for the moment I think it would be good to have a sort of collection of customFunctions that people can use and adapt to their needs. It may be an actual separate repo, now that I think about it.

rs028:
I have also changed the custom function you wrote for the CI testing, for consistency. So we use `t` for time and `temp` for temperature. 
Fortran is case insensitive, but I don't think there is a need to set both `t` and `T` as reserved "words" in the python script (if it turns out there is, we can always change it back).

AlfredMayhew:
All sounds good to me! Thanks.

Issue #414, opened 2020-02-08
shared library directory


rs028:
As it is now, the build script always creates `mechanism.so` in `model/configuration`. This makes the `--shared-lib` flag of the executable basically useless (although it still needs to be explicitly set if `--model` or `--configuration` are not set to the default values).

This is not a big issue, but it would make sense to

1. reintroduce the argument for the shared library in `build.sh`

OR

1. remove the `--shared_lib` flag (the easiest way would be to just hardcode it, like for `reactionRates_dir`)

spco:
Unless I am misunderstanding, this is not quite right. `build/build_atchem2.sh` has a second optional argument which indicates the directory to place `mechanism.f90` and `mechanism.so` in (as well as `mechanism.species` etc - this is also passed as the second argument to `build/mech_converter.py`. This defaults to `model/configuration/`, but can be anywhere. So I believe the first option above is already the case. `--shared_lib` then follows that case of defaulting to the same location.

rs028:
Sorry, that was poorly phrased.  What I meant is that there is no difference between the location of the shared library and that of the other `mechanism.*` files. The script takes 3 arguments. The second argument is the same for both the `build/mech_converter.py` and `make sharedlib` commands.

So, this is fine:

    ./atchem2 --configuration=project/config --shared_lib=project/config

but, this won't work:

    ./atchem2 --configuration=project/config --shared_lib=project/mech

Now the question is whether it makes sense to have the shared library separated by the other config files or not. In the former case we might as well remove the `--shared_lib` switch from the executable, in the latter we need to add a separate argument in `build.sh` for the `make sharedlib` command.

PS: if we are reworking the `build.sh` script we might consider #416 as well.

mixtli-c:
FWIW, it can cause some confusion when building AtChem for different projects following Figure 4.1 in the manual. If the shared_lib flag is not set, it throws a fortran error that is hard to parse by an inexperienced user (e.g. me)

PR #535, opened 2025-07-10
Set shared library directory

rs028:
Solves a couple of issues with the shared library.

One is described in #414: the build script uses the same argument as destination for the fortran mechanism files and for the shared library, even though (theoretically) the shared library can be elsewhere (as set by the `$SHAREDLIBDIR` variable in the Makefile). This makes the `--shared_lib` flag for the executable redundant. I think it makes sense for the shared library to be always in the same folder as the other mechanism files, especially now that we have a `customRateFuncs.f90` in the same directory. With this PR we force `mechanism.so` to always be in the configuration directory and remove the associated flag from the executable. The underlying structure remains in place, so we can always reintroduce the flag in the future, if we change our minds.

The second issue is related to the conversion of the `.fac` file. The build script always runs the python converter, which means that the fortran mechanism files are always updated, even if they are not changed. This defies the purpose of having a shared library, which should only be recompiled if the `.fac` file is changed. To correct this, the execution of the python script is moved from the build script to the Makefile. The build script is simplified but the user interface does not change.

Also introduces some error handling and more comments for both the Makefile and the build script.

### DIFF
--- a/tools/install/Makefile.skel
+++ b/tools/install/Makefile.skel
@@ -140,9 +140,9 @@ $(MECH_GEN): $(MECHFILE) $(MECH_CONVERTER)
 # the shared library is rebuilt if the mechanism fortran files (`mechanism.*`)
 # in $MECHDIR or the `customRateFuncs.f90` file in $CONFIGDIR are modified
 $(SHAREDLIB): $(MECH_GEN)
-	$(FORT_COMP) -c $(SRC)/dataStructures.f90 $(FSHAREDFLAGS) -o $(MECHDIR)/dataStructures.o -J$(OBJ) -I$(OBJ)
+	$(FORT_COMP) -c $(SRC)/dataStructures.f90 $(FSHAREDFLAGS) -o $(MECHDIR)/dataStructures.o -J$(MECHDIR) -I$(MECHDIR)
 	$(FORT_COMP) -c $(CONFIGDIR)/customRateFuncs.f90 $(FSHAREDFLAGS) -o $(MECHDIR)/customRateFuncs.o -J$(MECHDIR) -I$(MECHDIR)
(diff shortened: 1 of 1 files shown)


rs028:
The auto-generated mechanism files (`mechanism.*` fortran files, `customRateFuncs.o`, and the shared library `mechanism.so`) are now created by default in the `model/include` directory. The content of `model/include` should be automatically rewritten every time the  `.fac` file or `customRateFuncs.f90` are modified (unless I messed up my `Makefile` instructions).

The `model/configuration` directory only contains the `*.config` and `*.parameters` files, plus `customRateFuncs.f90`.

The `.fac` file can be wherever the user wants: by default, it is in `model/` although it may make more sense for it to be in `model/configuration`. Would be good to get some user's opinions on this.

rs028:
It may be my mistake. In the Makefile we have:

```
$(SHAREDLIB): $(MECH_GEN)
	$(FORT_COMP) -c $(SRC)/dataStructures.f90 $(FSHAREDFLAGS) -o $(MECHDIR)/dataStructures.o -J$(OBJ) -I$(OBJ)
	$(FORT_COMP) -c $(CONFIGDIR)/customRateFuncs.f90 $(FSHAREDFLAGS) -o $(MECHDIR)/customRateFuncs.o -J$(MECHDIR) -I$(MECHDIR)
	$(FORT_COMP) -c $(MECHDIR)/mechanism.f90 $(FSHAREDFLAGS) -o $(MECHDIR)/mechanism.o -J$(OBJ) -I$(OBJ)
	$(FORT_COMP) -shared -o $(SHAREDLIB) $(MECHDIR)/dataStructures.o $(MECHDIR)/customRateFuncs.o $(MECHDIR)/mechanism.o
```

Try changing `-J$(OBJ) -I$(OBJ)` to `-J$(MECHDIR) -I$(MECHDIR)` and see if it solves the issue. These tell the compiler the destination of the modules and where to find them.

PR #548, opened 2026-01-19
Move and rename the include directory to sharedlib

rs028:
This is the final step in reorganizing the `model/` directory. The new structure is:

- `configuration` : the  configuration directory, set by build script flag `--configuration=`. Includes the `*.config` and `*.parameters` files, plus `customRateFuncs.f90`. These files are editable by the user.
- `constraints` : the constraints directory (not changed)
- `output` : the output directory (not changed)
- `sharedlib` : the shared library directory, set by build script flag `--shared_lib=` . Includes the auto-generated `mechanism.*` files and the pre-compiled mechanism shared library (`mechanism.so`). These files should not be editable by the user.
- `mechanism.fac` : the chemical mechanism file, set by build script flag `--mechanism=`, During build, this is converted in the `mechanism.*` files and compiled together with `customRateFuncs.f90` into the mechanism shared library.

Other changes:
- the build script flag `--mcm=` is repurposed to indicate the version of the MCM files
- the executable does not have a `--mcm=` flag anymore
- the executable flag `--shared_lib` does not point to the `mechanism.so` file, but to the directory where the file is located (i.e. `model/sharedlib`

@AlfredMayhew @bsn502, any comments?


AlfredMayhew (approved):
I think you're right about the tab completion being to do with my set-up. I'm running with Windows Subsystem for Linux, so it get's a bit weird sometimes.
I've given this a quick go and it seems to work as expected. Thanks!

